PR #6, opened 2022-05-06
add a default version in case we cannot get it from build args

mattkim:
## What was changed
add a default version in case we cannot get it from build args

## Why?
so that we can still got a version even when tcld is not installed using build args

### DIFF
--- a/app/version.go
+++ b/app/version.go
@@ -2,6 +2,10 @@ package app
 
 import "github.com/urfave/cli/v2"
 
+const (
+	DefaultVersion = "0.1.1"
+)
+
 var (
 	BuildDate string
 	Commit    string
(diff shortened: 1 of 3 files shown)


app/version.go
```diff
@@ -2,6 +2,10 @@ package app
 
 import "github.com/urfave/cli/v2"
 
+const (
+	DefaultVersion = "v0.1.1"
```
anekkanti, line 6: Why not just set the Version variable with this default?

mattkim: I just tried this--and it doesn't work, looks like the value in the cli always takes precedent. I thought the build arg to take precedent. It is a little more reliable I suppose since it is automated.

anekkanti (approved):

PR #11, opened 2022-05-09
update installation to include homebrew and bump version

mattkim:
## What was changed
update installation to include homebrew and bump version

## Why?
we can install tcld with homebrew now

1. How was this tested:
unit tests

3. Any docs updates needed?
yes

### DIFF
--- a/app/version.go
+++ b/app/version.go
@@ -3,7 +3,7 @@ package app
 import "github.com/urfave/cli/v2"
 
 const (
-	DefaultVersion = "v0.1.1"
+	DefaultVersion = "v0.1.2"
 )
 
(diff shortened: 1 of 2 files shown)


app/version.go
```diff
@@ -3,7 +3,7 @@ package app
 import "github.com/urfave/cli/v2"
 
 const (
-	DefaultVersion = "v0.1.1"
+	DefaultVersion = "v0.1.2"
```
LakshmiRavali, line 6: Is this the upcoming version?

mattkim: yes I will tag the commit right after

PR #18, opened 2022-05-20
add list/add/remove ca certs commands for better user experience. 

anekkanti:


app/ca.go
```diff
@@ -0,0 +1,117 @@
...
+	// decode the cert bundle
+	bundle, err := base64.StdEncoding.DecodeString(encodedCerts)
+	if err != nil {
+		return nil, err
+	}
+	if len(bundle) == 0 {
+		return caCerts{}, nil
+	}
+
+	certBlocks := [][]byte{}
```
mastermanu, line 36: [Nit] extra lines 36-57 into a separate helper function

app/ca.go
```diff
@@ -0,0 +1,117 @@
...
+	p := bundle
+	for {
+		var block *pem.Block
+		var rem []byte
+		block, rem = pem.Decode(p)
+		if block == nil {
+			break
+		}
+		blocks = append(blocks, block.Bytes...)
+		certBlocks = append(certBlocks, []byte(strings.TrimSpace(string(p[:len(p)-len(rem)]))))
```
mastermanu, line 47: just to clarify, the purpose of this was to capture any optional headers that could have been supplied?

mastermanu: This also includes the BEGIN CERTIFICATE / END CERTIFICATE, whereas block contains just the bytes within those demarcations?

app/ca.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	if len(blocks) == 0 {
+		return nil, fmt.Errorf("failed to decode certificates")
+	}
+	// If p is greater than 0, then this means that there was a portion of the certificate that
+	// is/was malformed.
+	if len(p) > 0 {
+		return nil, fmt.Errorf("failed to parse one or more certificates, remainingBytesLen=%d", len(p))
+	}
+	certs, err := x509.ParseCertificates(blocks)
```
mastermanu, line 58: this data comes from the server, right? it's fine to be safe, but am wondering when we might hit this situation

app/ca.go
```diff
@@ -0,0 +1,117 @@
...
+			break
+		}
+		blocks = append(blocks, block.Bytes...)
+		certBlocks = append(certBlocks, []byte(strings.TrimSpace(string(p[:len(p)-len(rem)]))))
+		p = rem
+	}
+	if len(blocks) == 0 {
+		return nil, fmt.Errorf("failed to decode certificates")
+	}
+	// If p is greater than 0, then this means that there was a portion of the certificate that
```
mastermanu, line 53: the server actually will still save the bundle even if there are leftover bytes from the bundle. We might want to print a warning here vs actually failing

app/ca.go
```diff
@@ -0,0 +1,117 @@
...
+	// is/was malformed.
+	if len(p) > 0 {
+		return nil, fmt.Errorf("failed to parse one or more certificates, remainingBytesLen=%d", len(p))
+	}
+	certs, err := x509.ParseCertificates(blocks)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse certificates: %s", err.Error())
```
mastermanu, line 60: with Errorf, you can use %w and err

app/ca.go
```diff
@@ -0,0 +1,117 @@
...
+		out[i].Base64EncodedData = base64.StdEncoding.EncodeToString(certBlocks[i])
+		out[i].Issuer = certs[i].Issuer.String()
+		out[i].Subject = certs[i].Subject.String()
+		out[i].NotBefore = certs[i].NotBefore
+		out[i].NotAfter = certs[i].NotAfter
+	}
+	return out, nil
+}
+
+func (c *caCerts) bundle() (string, error) {
```
mastermanu, line 75: [Nit]. caCerts is already a slice. A pointer seems unnecessary here

app/ca.go
```diff
@@ -0,0 +1,117 @@
...
+	out := make([]byte, 0)
+	for i := range *c {
+		decodedCert, err := base64.StdEncoding.DecodeString((*c)[i].Base64EncodedData)
```
mastermanu, line 78: [Nit] decodedCert --> certBytes

app/ca.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		if len(out) != 0 {
+			out = append(out, []byte("\n")...)
+		}
+		out = append(out, decodedCert...)
+	}
+	return base64.StdEncoding.EncodeToString(out), nil
+}
+
+func (c *caCerts) add(certs caCerts) error {
```
mastermanu, line 90: [Nit] change signature to return (caCerts, error)

app/ca.go
```diff
@@ -0,0 +1,117 @@
...
+			if certs[i].Fingerprint == (*c)[j].Fingerprint {
+				return fmt.Errorf("certificate with fingerprint '%s' already exists", (*c)[j].Fingerprint)
+			}
+		}
+	}
+	*c = append(*c, certs...)
+	return nil
+}
+
+func (c *caCerts) remove(certs caCerts) error {
```
mastermanu, line 103: [Nit] change signature to return (caCerts, error)

app/ca.go
anekkanti: Yes that is correct. This was to keep everything in the PEM file as is.

app/namespace.go
```diff
@@ -14,6 +14,24 @@ import (
...
 )
 
+const (
+	caCertificateFlagName     = "ca-certificate"
+	caCertificateFileFlagName = "ca-certificate-file"
+)
+
+var (
+	caCertificateFlag = &cli.StringFlag{
+		Name:    "ca-certificate",
```
mastermanu, line 24: [Nit] use the constant defined above?

app/namespace.go
```diff
@@ -14,6 +14,24 @@ import (
...
+		Usage:   "the base64 encoded ca certificate",
+		Aliases: []string{"c"},
+	}
+	caCertificateFileFlag = &cli.PathFlag{
+		Name:    "ca-certificate-file",
```
mastermanu, line 29: [Nit] use the constant defined above?

app/namespace.go
```diff
@@ -14,6 +14,24 @@ import (
...
+	caCertificateFileFlagName = "ca-certificate-file"
+)
+
+var (
+	caCertificateFlag = &cli.StringFlag{
+		Name:    "ca-certificate",
+		Usage:   "the base64 encoded ca certificate",
+		Aliases: []string{"c"},
+	}
+	caCertificateFileFlag = &cli.PathFlag{
```
mastermanu, line 28: nice, I did not know there was a PathFlag

app/namespace.go
```diff
@@ -65,34 +83,43 @@ func (c *NamespaceClient) getNamespace(namespace string) (*ns.Namespace, error)
...
 }
 
-func (c *NamespaceClient) updateNamespace(
-	namespace string,
-	requestID string,
-	resourceVersion string,
-	spec *ns.NamespaceSpec,
-) error {
+func (c *NamespaceClient) updateNamespace(ctx *cli.Context, n *ns.Namespace) error {
+	if n.State != ns.STATE_ACTIVE {
```
mastermanu, line 87: should we rely on server for this kind of validation instead of putting this here?

app/namespace.go
```diff
@@ -138,179 +182,215 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+				Aliases: []string{"r"},
+				Flags: []cli.Flag{
+					NamespaceFlag,
+					RequestIDFlag,
+					ResourceVersionFlag,
+					caCertificateFlag,
+					caCertificateFileFlag,
+				},
+				Action: func(ctx *cli.Context) error {
+					cert, err := readCACerts(ctx)
```
mastermanu, line 259: most of this code seems super-common between this and above. factoring?

app/namespace.go
```diff
@@ -138,179 +182,215 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+					if err != nil {
+						return err
+					}
+					n, err := c.getNamespace(ctx.String(NamespaceFlagName))
+					if err != nil {
+						return err
+					}
+					if n.Spec.AcceptedClientCa == cert {
```
mastermanu, line 309: will this capture any changes in cert ordering

app/namespace.go
```diff
@@ -138,179 +182,215 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+						return err
+					}
+					n, err := c.getNamespace(ctx.String(NamespaceFlagName))
+					if err != nil {
+						return err
+					}
+					if n.Spec.SearchAttributes == nil {
+						n.Spec.SearchAttributes = make(map[string]namespace.NamespaceSpec_SearchAttributeType)
+					}
+					for attrName, attrType := range csa {
```
mastermanu, line 347: same question here: can we rely on the server

app/ca.go
anekkanti: The server will reject the bundle if it cannot parse it right? I believe we perform this exact check on the server.

app/ca.go
anekkanti: This code path is used to parse certs coming from the server as well as certs provided by the user.

app/ca.go
mastermanu: note that the server got is doing fmt.Println, not returning an error, whereas your code would return an error.

mastermanu (dismissed):

app/ca.go
anekkanti: updated as suggested.

app/ca.go
anekkanti: fixed.

app/ca.go
anekkanti: fixed.

app/ca.go
anekkanti: fixed

app/ca.go
anekkanti: fixed.

app/ca.go
anekkanti: fixed

app/namespace.go
anekkanti: fixed.

app/namespace.go
anekkanti: fixed.

app/namespace.go
anekkanti: You are right, we should...
Removed the check here.
Currently we do the check in the entity, but I will add the check on the service itself.

app/namespace.go
anekkanti: If the cert ordering changes, we will allow the update. Which seems fine... 
Really don't want to add logic to ignore order changes.

app/namespace.go
anekkanti: we can't, add is just a cli side concept. it relies on the update namespace op.

app/namespace.go
anekkanti: did some refactoring.

app/ca.go
anekkanti: Got it. 
I am wondering if we should be more strict on the cli, especially when they are adding and remove certs to not have malformed inputs. If they still want to push partial certs they can still use the `set` command.

mattkim (approved):

PR #32, opened 2022-06-18
Update api/temporalcloudapi

feihuang:
## What was changed

Update proto generated files

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


mastermanu (approved):

api/temporalcloudapi/namespace/v1/message.pb.go
```diff
@@ -41,23 +41,23 @@ const (
 )
 
 var NamespaceSpec_SearchAttributeType_name = map[int32]string{
-	0: "Unspecified",
```
mastermanu, line 44: this won't break CLI, right? CLI i think translates user provided flag values into the enum values.

feihuang: as discussed, removed these changes.

PR #43, opened 2022-06-27
Adds ability to configure certificate-filters on a namespace via the CLI

mastermanu:
Adds new namespace sub-command called certificate-filters ("cf"). This sub-commands has 3 sub-commands:

import/imp --> imports certificate filter configuration from either a file or from input on the CLI
export/exp --> takes the existing certificate filter configuration and writes it to the output (and also optionally to a file)
clear --> removes all certificate filters on the namespace.

import/clear will print out the before/after of the certificate-filter configuration to make it more obvious to the user what is being changed.

```
tcld n
NAME:
   tcld namespace - Namespace operations

USAGE:
   tcld namespace command [command options] [arguments...]

COMMANDS:
   list, l                  List all known namespaces
   get, g                   Get namespace information
   accepted-client-ca, ca   Manage client ca certificate used to verify client connections
   certificate-filters, cf  Manage optional certificate filters used by namespace to authorize client certificates based on distinguished name fields
   search-attributes, sa    Manage search attributes used by namespace
   help, h                  Shows a list of commands or help for one command
```

```
tcld n cf
NAME:
   tcld namespace certificate-filters - Manage optional certificate filters used by namespace to authorize client certificates based on distinguished name fields

USAGE:
   tcld namespace certificate-filters command [command options] [arguments...]

COMMANDS:
   import, imp  Sets the certificate filters on the namespace
   export, exp  Exports existing certificate filters on the namespace
   clear, c     Clears all certificate filters on the namespace. Note that this will allow *any* client certificate that chains up to a configured CA in the bundle to connect to the namespace
   help, h      Shows a list of commands or help for one command

OPTIONS:
   --help, -h  show help (default: false)
```

```
tcld n cf imp
NAME:
   tcld namespace certificate-filters import - Sets the certificate filters on the namespace

USAGE:
   tcld namespace certificate-filters import [command options] [arguments...]

OPTIONS:
   --certificate-filter-file value, --file value, -f value    Path to a JSON file that defines the certificate filters that will be configured on the namespace. This will replace the existing filter configuration. Sample JSON: { "filters": [ { "commonName": "test1" } ] }
   --certificate-filter-input value, --input value, -i value  JSON that defines the certificate filters that will be configured on the namespace. This will replace the existing filter configuration. Sample JSON: { "filters": [ { "commonName": "test1" } ] }
   --namespace value, -n value                                The namespace hosted on temporal cloud [$TEMPORAL_CLOUD_NAMESPACE]
   --request-id value, -r value                               The request-id to use for the asynchronous operation, if not set the server will assign one (optional)
   --resource-version value, -v value                         The resource-version (etag) to update from, if not set the cli will use the latest (optional)

```

```
tcld n cf exp
NAME:
   tcld namespace certificate-filters export - Exports existing certificate filters on the namespace

USAGE:
   tcld namespace certificate-filters export [command options] [arguments...]

OPTIONS:
   --certificate-filter-file value, --file value, -f value  Path to a JSON file where tcld will export the certificate filter configuration to
   --namespace value, -n value                              The namespace hosted on temporal cloud [$TEMPORAL_CLOUD_NAMESPACE]
   --request-id value, -r value                             The request-id to use for the asynchronous operation, if not set the server will assign one (optional)
   --resource-version value, -v value                       The resource-version (etag) to update from, if not set the cli will use the latest (optional)
   
```

```
tcld n cf clear
NAME:
   tcld namespace certificate-filters clear - Clears all certificate filters on the namespace. Note that this will allow *any* client certificate that chains up to a configured CA in the bundle to connect to the namespace

USAGE:
   tcld namespace certificate-filters clear [command options] [arguments...]

OPTIONS:
   --namespace value, -n value         The namespace hosted on temporal cloud [$TEMPORAL_CLOUD_NAMESPACE]
   --request-id value, -r value        The request-id to use for the asynchronous operation, if not set the server will assign one (optional)
   --resource-version value, -v value  The resource-version (etag) to update from, if not set the cli will use the latest (optional)

```


app/cafilters.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type certificateFilter struct {
+	CommonName             string `json:"commonName"`
+	Organization           string `json:"organization"`
+	OrganizationalUnit     string `json:"organizationalUnit"`
+	SubjectAlternativeName string `json:"subjectAlternativeName"`
+}
+
+type certificateFiltersConfig struct {
```
mastermanu, line 19: open to having this just be a top-level array vs an array embedded in a field. thoughts here?

app/namespace.go
```diff
@@ -165,277 +167,427 @@ func ReadCACerts(ctx *cli.Context) (string, error) {
...
-					if _, exists := n.Spec.SearchAttributes[newName]; exists {
-						return fmt.Errorf("search attribute with new name '%s' already exists", ctx.String("new-name"))
-					}
-					y, err := ConfirmPrompt(ctx, "renaming search attribute may cause failures if any worker is still using the old name of the search-attributes. confirm rename?")
-					if err != nil || !y {
-						return err
-					}
-					return c.renameSearchAttribute(ctx, n, existingName, newName)
+				{
+					Name:    "certificate-filters",
```
mastermanu, line 345: this can be nested within accepted-client-ca instead of being its own top-level thing. thoughts?

feihuang (approved):

app/namespace.go
anekkanti: I vote to keep them as is.

app/namespace.go
```diff
@@ -165,277 +167,427 @@ func ReadCACerts(ctx *cli.Context) (string, error) {
...
+								}
+
+								fmt.Println("certificate filters before change:")
+								if err := PrintObj(fromSpec(n.Spec.CertificateFilters)); err != nil {
+									return err
+								}
+
+								n.Spec.CertificateFilters = filters.toSpec()
+
+								fmt.Println("certificate filters after change:")
```
anekkanti, line 407: Do we want to maybe send a confirmation prompt before the user commits to the update? like we do for ca cert updates?

app/cafilters.go
anekkanti: I think this is better. Gives us an option to put more information in the object if the need ever arises.

app/namespace.go
```diff
@@ -165,277 +167,427 @@ func ReadCACerts(ctx *cli.Context) (string, error) {
...
+								if err != nil {
+									return err
+								}
+
+								fmt.Println("certificate filters to be cleared:")
+								if err := PrintObj(fromSpec(n.Spec.CertificateFilters)); err != nil {
+									return err
+								}
+
+								n.Spec.CertificateFilters = nil
```
anekkanti, line 475: We should throw a conformation prompt here too.

anekkanti (approved):
looks good, just a few minor comments.

app/namespace.go
mastermanu: yes, i was weighing doing this, should have asked that as an open question. I will add confirmation and an optional --force flag?

PR #54, opened 2022-08-02
Switch api/ to protogen/

hehaifengcn:
## Why?

so we can reuse `make proto` command from saas-control-plane Makefile (with minor tweaks) to generate proto go files.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


feihuang:
LGTM

anekkanti (approved):

PR #55, opened 2022-08-03
Adds ability to configure account metrics via tcld commands

mastermanu:


app/account.go
```diff
@@ -0,0 +1,300 @@
...
+						{
+							Name:  "enable",
+							Usage: "Enables the metrics endpoint. CA Certificates *must* be configured prior to enabling the endpoint",
+							Action: func(ctx *cli.Context) error {
+								a, err := c.getAccount()
+								if err != nil {
+									return err
+								}
+
+								if a.Spec.Metrics != nil && a.Spec.Metrics.EnableMetricsEndpoint {
```
mastermanu, line 119: Will change this proto to just be "enable"

app/account.go
```diff
@@ -0,0 +1,300 @@
...
+									return errors.New("metrics endpoint is already enabled")
+								}
+
+								if a.Spec.Metrics == nil || a.Spec.Metrics.AcceptedMetricsClientCa == "" {
```
mastermanu, line 123: will change this proto to just AcceptedClientCa

protogen/api/auditlog/v1/message.pb.go
```diff
@@ -0,0 +1,1736 @@
+// Code generated by protoc-gen-gogo. DO NOT EDIT.
+// source: api/auditlog/v1/message.proto
+
+package auditlog
```
underrun, line 4: why is the auditlog proto go file added as part of this?

mastermanu: because it came with syncing to the latest protos

underrun (approved):

PR #70, opened 2022-08-31
add support for optional sync flag to block on async ops till completion

anekkanti:


app/request.go
```diff
@@ -2,12 +2,37 @@ package app
...
 
+const (
+	SyncRequestFlagName    = "synchronous-request"
+	RequestTimeoutFlagName = "request-timeout"
+)
+
+var (
+	RequestTimeoutFlag = &cli.DurationFlag{
+		Name:    RequestTimeoutFlagName,
+		Usage:   "Time to wait for asynchronous requests to finish",
```
mastermanu, line 23: how does this flag work? As in what values are expected?

Is it easier to have this be something like "timeout in seconds" ?

app/request.go
```diff
@@ -66,8 +91,102 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
...
-				return c.getRequestStatus(ctx.String("request-id"))
+				res, err := c.getRequestStatus(ctx, ctx.String("request-id"))
+				if err != nil {
+					return err
+				}
+				return PrintProto(res)
+			},
+		}, {
+			Name:    "wait",
+			Usage:   "wait till the request completes",
```
mastermanu, line 102: [Nit] "wait for the request to complete"

app/request.go
```diff
@@ -66,8 +91,102 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
...
+				return err
+			}
+			switch status.State {
+			case request.STATE_FULFILLED:
+				break loop
+			case request.STATE_FAILED:
+				fmt.Fprintf(writer, "operation failed \n")
+				return fmt.Errorf("request failed: %s", status.FailureReason)
+			case request.STATE_CANCELLED:
+				fmt.Fprintf(writer, "operation failed \n")
```
mastermanu, line 156: [Nit] operation canceled

app/request.go
```diff
@@ -66,8 +91,102 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
...
+) error {
+
+	if ctx.Bool(SyncRequestFlagName) {
+		return c.waitOnRequest(ctx, operation, status.RequestId)
+	}
+
+	if err := PrintProto(status); err != nil {
+		return err
+	}
+	fmt.Sprintf("started %s operation with requestId='%s', to monitor its progress use command: `%s request get -r '%s'`", operation, status.RequestId, AppName, status.RequestId)
```
mastermanu, line 190: [Nit] is this going to break anybody who has written scripts based on the output of commands from before? That could be annoying, maybe we don't print this to standard out or we have an option to suppress this?

mastermanu (approved):

app/request.go
anekkanti: It uses the string to duration func in the time package: https://pkg.go.dev/time#ParseDuration
You can pass in values like: 10s, 1h, 1h20m...

app/request.go
```diff
@@ -2,12 +2,37 @@ package app
...
+	RequestTimeoutFlag = &cli.DurationFlag{
+		Name:    RequestTimeoutFlagName,
+		Usage:   "Time to wait for asynchronous requests to finish",
+		EnvVars: []string{"REQUEST_TIMEOUT"},
+		Aliases: []string{"rt"},
+		Value:   time.Hour,
+	}
+	SyncRequestFlag = &cli.BoolFlag{
+		Name:    SyncRequestFlagName,
+		Usage:   "Block on request to complete",
```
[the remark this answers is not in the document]
anekkanti: Moved all mentions to wait.

app/request.go
anekkanti: fixed

app/request.go
anekkanti: fixed

app/request.go
```diff
@@ -30,22 +55,22 @@ func GetRequestClient(ctx *cli.Context) (*RequestClient, error) {
...
+	return res.RequestStatus, nil
 }
 
 func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error) {
 
 	var c *RequestClient
 	return CommandOut{Command: &cli.Command{
 		Name:    "request",
-		Usage:   "Manage asynchronous requests",
+		Usage:   "Manage requests",
```
tminusplus, line 73: I wonder if `long-running requests` could more clear for the user.

app/request.go
```diff
@@ -60,14 +85,108 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
 			Flags: []cli.Flag{
 				&cli.StringFlag{
 					Name:     "request-id",
-					Usage:    "The request-id of the asynchronous request",
+					Usage:    "The request-id of the request",
```
tminusplus, line 88: `The request's ID` could be less circular, or `Unique identifier of the request`

app/request.go
```diff
@@ -60,14 +85,108 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
...
+					return err
+				}
+				return PrintProto(res)
+			},
+		}, {
+			Name:    "wait",
+			Usage:   "wait for the request complete",
+			Aliases: []string{"w"},
+			Flags: []cli.Flag{
+				&cli.StringFlag{
```
tminusplus, line 105: We could define this as a var since it is a repeated flag

app/request.go
```diff
@@ -60,14 +85,108 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
...
+				return c.waitOnRequest(ctx, "", ctx.String("request-id"))
 			},
 		}},
 	}}, nil
 }
+
+func (c *RequestClient) waitOnRequest(ctx *cli.Context, operation string, requestID string) error {
+
+	ticker := time.NewTicker(time.Millisecond)
+	timer := time.NewTimer(ctx.Duration(RequestTimeoutFlagName))
```
tminusplus, line 122: Defer calls for the ticker and timer are missing.

app/request.go
```diff
@@ -60,14 +85,108 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
...
+				return fmt.Errorf("request was cancelled: %s", status.FailureReason)
+			}
+			if operation != "" {
+				fmt.Fprintf(writer, "waiting for %s operation (requestId='%s') to finish, current state: %s\n",
+					operation, requestID, request.State_name[int32(status.State)])
+			} else {
+				fmt.Fprintf(writer, "waiting for request with id='%s' to finish, current state: %s\n",
+					requestID, request.State_name[int32(status.State)])
+			}
+			ticker.Reset(time.Second * time.Duration(status.CheckDuration.Seconds))
```
tminusplus, line 166: If `status.CheckDuration.Seconds` is 0 `ticker.Reset` [will panic](https://pkg.go.dev/time#Ticker.Reset). Should we set a minimum value to ensure that this never happens and ensure that the ticker fires less often than 1ms?

app/account.go
```diff
@@ -147,7 +164,12 @@ func NewAccountCommand(getAccountClientFn GetAccountClientFn) (CommandOut, error
 								}
 
 								a.Spec.Metrics.Enabled = false
-								return c.updateAccount(ctx, a)
+								status, err := c.updateAccount(ctx, a)
+								if err != nil {
+									return err
+								}
+								return r.HandleRequestStatus(ctx, "disable metrics", status)
+
```
tminusplus, line 172: nit: extra space

app/request.go
tminusplus: +1 we could check if stdout `isatty` and also have an env flag for disabling this (non-interactive mode maybe?)

anekkanti: I like the idea to use isatty, to disable such messages. 
I think we can leverage such mechanism to hide other messaging we do. 

Let me address these in the next PR.

app/request.go
anekkanti: `long-running` seems technically wrong, because some write operations we do will be very quick.

app/request.go
anekkanti: fixed.

app/request.go
anekkanti: Good catch, fixed.

app/request.go
anekkanti: thanks, fixed.

app/request.go
anekkanti: actually we should probably change this to operation id. to keep things in consistent with the new apis
will address this in the next PR

tminusplus: That makes sense

app/request.go
tminusplus: Async operation makes sense as it is inline with the new APIs

app/prompt.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	AutoConfirmFlagName = "auto_confirm"
+	AutoConfirmFlagName = "auto-confirm"
```
tminusplus, line 13: I think should keep this as `auto_confirm` since folks are using it now. That or we could print a helpful message asking them to use auto-confirm instead - but that could break some use-cases folks have.

app/request.go
```diff
@@ -59,15 +86,122 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
...
+			if err := PrintProto(status); err != nil {
+				fmt.Fprintf(writer, "failed to print status: %s", err)
+			}
+		}
+	}()
+loop:
+	for {
+		select {
+		case <-timer.C:
+			return fmt.Errorf("timed out waiting for request to complete, namespace=%s, requestID=%s, timeout=%s",
```
tminusplus, line 144: For some requests namespace could be empty right? If so I think we should conditionally add the field.

app/request.go
```diff
@@ -59,15 +86,122 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
...
+				break loop
+			case request.STATE_FAILED:
+				fmt.Fprintf(writer, "operation failed \n")
+				return fmt.Errorf("request failed: %s", status.FailureReason)
+			case request.STATE_CANCELLED:
+				fmt.Fprintf(writer, "operation cancelled\n")
+				return fmt.Errorf("request was cancelled: %s", status.FailureReason)
+			}
+			if operation != "" {
+				fmt.Fprintf(writer, "waiting for %s operation (requestId='%s') to finish, current state: %s\n",
```
tminusplus, line 166: `id='%s'` to keep it consistent with the other log statements?

app/request.go
```diff
@@ -59,15 +86,122 @@ func NewRequestCommand(getRequestClientFn GetRequestClientFn) (CommandOut, error
...
+		// status can be empty when the operation is cancelled
+		return nil
+	}
+	if ctx.Bool(WaitForRequestFlagName) {
+		return c.waitOnRequest(ctx, operation, status.RequestId)
+	}
+	if err := PrintProto(status); err != nil {
+		return err
+	}
+	fmt.Printf(
```
tminusplus, line 202: I think we should allow these messages to be silenced for scripted use cases which expect JSON output on success. Perhaps `isatty` gets us this in the follow-up PR? Unsure if it covers all cases or not.

tminusplus (approved):

app/prompt.go
anekkanti: added an alias to the flag, to support both.

app/request.go
anekkanti: yup, removed namespace.

app/request.go
anekkanti: fixed.

app/request.go
anekkanti: I plan on addressing this comprehensively in the next PR. 
In the next PR i hope to add a tcld version checker, which notifies the user to update their `tcld` when it gets old.

tminusplus (approved):

PR #93, opened 2022-10-26
Add retention flag, call to update retention

laniehei:
Add ability to update retention on a namespace

### DIFF
--- a/app/flags.go
+++ b/app/flags.go
@@ -33,7 +33,7 @@ var (
 		Hidden:   true,
 		Required: false,
 	}
-	RetentionFlag = &cli.StringFlag{
+	RetentionFlag = &cli.IntFlag{
 		Name:     RetentionDaysFlagName,
 		Usage:    "The retention of the namespace in days",
 		Aliases:  []string{"rd"},
--- a/app/flags.go
+++ b/app/flags.go
@@ -33,7 +33,7 @@ var (
 		Hidden:   true,
 		Required: false,
 	}
-	RetentionFlag = &cli.IntFlag{
+	RetentionDaysFlag = &cli.IntFlag{
 		Name:     RetentionDaysFlagName,
 		Usage:    "The retention of the namespace in days",
 		Aliases:  []string{"rd"},
--- a/app/flags.go
+++ b/app/flags.go
@@ -35,7 +35,7 @@ var (
 	}
 	RetentionFlag = &cli.StringFlag{
 		Name:     RetentionDaysFlagName,
-		Usage:    "The retention of the namespace",
+		Usage:    "The retention of the namespace in days",
 		Aliases:  []string{"rd"},
 		EnvVars:  []string{"NAMESPACE_RETENTION"},
 		Required: true,
--- a/app/flags.go
+++ b/app/flags.go
@@ -36,7 +36,7 @@ var (
 	RetentionFlag = &cli.StringFlag{
 		Name:     RetentionDaysFlagName,
 		Usage:    "The retention of the namespace",
-		Aliases:  []string{"ret"},
+		Aliases:  []string{"rd"},
 		EnvVars:  []string{"NAMESPACE_RETENTION"},
 		Required: true,
 	}
--- a/app/flags.go
+++ b/app/flags.go
@@ -10,7 +10,7 @@ import (
 const (
 	ServerFlagName          = "server"
 	ConfigDirFlagName       = "config-dir"
-	RetentionFlagName       = "retention-days"
+	RetentionDaysFlagName   = "retention-days"
 	NamespaceFlagName       = "namespace"
 	RequestIDFlagName       = "request-id"
 	ResourceVersionFlagName = "resource-version"
@@ -34,7 +34,7 @@ var (
 		Required: false,
 	}
 	RetentionFlag = &cli.StringFlag{
-		Name:     RetentionFlagName,
+		Name:     RetentionDaysFlagName,
 		Usage:    "The retention of the namespace",
 		Aliases:  []string{"ret"},
 		EnvVars:  []string{"NAMESPACE_RETENTION"},
--- a/app/flags.go
+++ b/app/flags.go
@@ -10,6 +10,7 @@ import (
 const (
 	ServerFlagName          = "server"
 	ConfigDirFlagName       = "config-dir"
+	RetentionFlagName       = "retention"
 	NamespaceFlagName       = "namespace"
 	RequestIDFlagName       = "request-id"
 	ResourceVersionFlagName = "resource-version"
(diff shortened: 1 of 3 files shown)


app/namespace.go
```diff
@@ -574,6 +574,28 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 						},
 					},
 				},
+				{
+					Name:    "retention",
```
feihuang, line 578: for consistency with other commands, we should do ocld namespace retention set --retention-days 30?

laniehei: done

app/flags.go
```diff
@@ -10,6 +10,7 @@ import (
 const (
 	ServerFlagName          = "server"
 	ConfigDirFlagName       = "config-dir"
+	RetentionFlagName       = "retention-days"
```
mastermanu, line 13: [Nit] RetentionDaysFlagName

app/flags.go
```diff
@@ -32,6 +33,13 @@ var (
 		Hidden:   true,
 		Required: false,
 	}
+	RetentionFlag = &cli.StringFlag{
+		Name:     RetentionFlagName,
+		Usage:    "The retention of the namespace",
+		Aliases:  []string{"ret"},
```
mastermanu, line 39: [Nit] tctl uses the alias "rd". Not sure if we want to align on that?
Same with the usage. tctl refers to it as the "workflow execution retention". Might be good to have both tcld and tctl use same description / verbiage if possible?

app/namespace.go
mastermanu: [Nit] given that we are specifying a flag on a subcommand called "retention", having "retention" in the flag name and the command seems redundant

app/namespace.go
```diff
@@ -574,6 +574,33 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+					Usage:   "The length of time a closed workflow will be preserved before deletion",
```
mastermanu, line 579: [Nit] re-word usage a bit as this is a command, not a flag (e.g. "Manages the length of time a closed workflow...")

app/namespace.go
```diff
@@ -574,6 +574,33 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+					Aliases: []string{"r"},
+					Subcommands: []*cli.Command{{
+						Name:    "set",
+						Aliases: []string{"s"},
+						Usage:   "Set the length of time a closed workflow will be preserved before deletion per namespace",
+						Flags: []cli.Flag{
```
mastermanu, line 585: shouldn't this take the ResourceVersion flag as well?

app/flags.go
mastermanu: Also, we should mention that the unit is days (even if the name of the flag makes it obvious, nice to also have the usage/help be explicit too) (e.g. "Workflow execution retention in days" or "The retention of the namespace in days")

app/namespace.go
```diff
@@ -574,6 +574,33 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+							RetentionFlag,
+							RequestIDFlag,
+						},
+						Action: func(ctx *cli.Context) error {
+							retention := ctx.Int(RetentionFlagName)
+							n, err := c.getNamespace(ctx.String(NamespaceFlagName))
+							if err != nil {
+								return err
+							}
+							if int32(retention) == n.Spec.RetentionDays {
```
mastermanu, line 596: [Nit] if there is no uint flag, might want to check that a negative value is not supplied

mastermanu (changes requested):
mostly minor comments

app/namespace.go
```diff
@@ -574,6 +574,33 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+								return fmt.Errorf("retention for namespace is already set at %d days", ctx.Int(RetentionFlagName))
+							}
+							n.Spec.RetentionDays = int32(retention)
+							return c.updateNamespace(ctx, n)
+						},
+					}},
+				},
```
mastermanu, line 603: Please add corresponding unit tests. See what we do for certificate filters, certificates and/or search attributes as an example

app/namespace.go
laniehei: I will keep this as "retention-days" to allow for usage of the same flag name

app/flags.go
laniehei: done

app/namespace.go
laniehei: done

app/namespace.go
```diff
@@ -574,6 +574,33 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 						},
 					},
 				},
+				{
+					Name:    "retention",
+					Usage:   "The length of time a closed workflow will be preserved before deletion",
+					Aliases: []string{"r"},
+					Subcommands: []*cli.Command{{
+						Name:    "set",
```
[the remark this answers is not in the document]
laniehei: done

app/namespace.go
laniehei: done

app/namespace.go
laniehei: uint was available thanks!

app/namespace.go
laniehei: added tests

app/flags.go
```diff
@@ -32,6 +33,13 @@ var (
 		Hidden:   true,
 		Required: false,
 	}
+	RetentionFlag = &cli.StringFlag{
```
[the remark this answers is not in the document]
laniehei: done

app/namespace.go
laniehei: I turned this into a uint flag, but it looks like it turns it into "0" if negative, so checking for negative.

app/namespace.go
```diff
@@ -574,6 +574,59 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+								return c.updateNamespace(ctx, n)
+							},
+						},
+						{
+							Name:    "get",
+							Aliases: []string{"g"},
+							Usage:   "Retrieve the length of time (in days) a closed workflow will be preserved before deletion for a given namespace",
+							Flags: []cli.Flag{
+								NamespaceFlag,
+								RequestIDFlag,
```
anekkanti, line 617: Don't think we need the request id flag for get command.

anekkanti (approved):

PR #97, opened 2022-11-01
add disable-pop-up flag + add logout command

mattkim:
## What was changed

add disable-pop-up flag + add logout command

## Why?

add logout command to log out of current session

add disable-pop-up flag for login/logout to allow user to pick which browser to open url

## Checklist

1. Closes 

2. How was this tested:

unit tests

3. Any docs updates needed?

### DIFF
--- a/app/logout.go
+++ b/app/logout.go
@@ -1,0 +1,38 @@
+package app
+
+import (
+	"fmt"
+	"github.com/urfave/cli/v2"
+)
+
+func NewLogoutCommand(c *LoginClient) (CommandOut, error) {
+	return CommandOut{Command: &cli.Command{
+		Name:    "logout",
+		Usage:   "Logout current user",
+		Aliases: []string{"lo"},
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:     "domain",
+				Value:    "login.tmprl.cloud",
+				Aliases:  []string{"d"},
+				Required: false,
+				Hidden:   true,
+			},
+			&cli.BoolFlag{
+				Name:     "disable-pop-up",
+				Usage:    "disable browser pop-up",
+				Required: false,
+			},
+		},
+		Action: func(ctx *cli.Context) error {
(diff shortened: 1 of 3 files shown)


app/logout.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func NewLogoutCommand(c *LoginClient) (CommandOut, error) {
+	return CommandOut{Command: &cli.Command{
+		Name:    "logout",
+		Usage:   "Logout current user",
+		Aliases: []string{"lo"},
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:     "domain",
```
feihuang, line 15: can we reuse the same flag in login.go?

app/logout.go
```diff
@@ -0,0 +1,38 @@
...
+				Value:    "login.tmprl.cloud",
+				Aliases:  []string{"d"},
+				Required: false,
+				Hidden:   true,
+			},
+			&cli.BoolFlag{
```
feihuang, line 21: same as above

app/logout.go
```diff
@@ -0,0 +1,38 @@
...
+				Name:     "disable-pop-up",
+				Usage:    "disable browser pop-up",
+				Required: false,
+			},
+		},
+		Action: func(ctx *cli.Context) error {
+			logoutURL := fmt.Sprintf("https://%s/v2/logout", ctx.String("domain"))
+			fmt.Printf("Logout via this url: %s\n", logoutURL)
+			if !ctx.Bool("disable-pop-up") {
```
feihuang, line 30: does it ever make sense to pop up?

feihuang (approved):

app/logout.go
mattkim: done

app/logout.go
mattkim: done

app/logout.go
mattkim: Yes, i believe pop-up is a good default behavior so you don't have to click on a link

PR #105, opened 2022-11-21
Fix small typo in error message when a CA cert does not exist

protochron:
## What was changed
Fixed a small typo in the error message the CLI outputs when a CA cert does not exist.

## Why?
See above. Typos aren't fun to have in CLI tools.


mattkim (approved):

tminusplus:
Thank you for the fix!

PR #151, opened 2023-03-10
add create namespace command

mattkim:
## What was changed

add create namespace command

## Why?

allow users to create a namespace using `tcld`

## Checklist

- feature parity with create namespace in `cloud.temporal.io`
- supports creating namespace with:
  - name
  - region
  - retention (in days)
  - ca-certificate
  - certificate-filters
  - search attributes
  - user namespace permissions

1. How was this tested:

unit tests


app/namespace.go
```diff
@@ -184,6 +272,129 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+							Required: true,
+						},
+						&cli.IntFlag{
+							Name:    RetentionDaysFlagName,
+							Usage:   "The retention of the namespace in days",
+							Aliases: []string{"rd"},
+							Value:   30,
+						},
+						&cli.StringFlag{
+							Name:    CaCertificateFlagName,
```
anekkanti, line 296: There is an existing flag that we could reuse: https://github.com/temporalio/tcld/blob/main/app/namespace.go#L26

app/namespace.go
```diff
@@ -5,24 +5,45 @@ import (
...
 )
 
 var (
+	namespaceRegionFlag = &cli.StringFlag{
+		Name:     namespaceRegionFlagName,
+		Usage:    "Create namespace in this region",
+		Aliases:  []string{"re"},
+		Required: true,
+	}
+	namespaceEnvironmentFlag = &cli.StringFlag{
```
mastermanu, line 40: what's the point of even adding this if its hidden right now? I believe server will auto-assign the value if its not set anyway, right?

app/namespace.go
```diff
@@ -33,23 +54,30 @@ var (
...
 	caCertificateFingerprintFlag = &cli.StringFlag{
 		Name:    caCertificateFingerprintFlagName,
 		Usage:   "The fingerprint of to the ca certificate",
 		Aliases: []string{"fp"},
 	}
+	searchAttributeFlag = &cli.StringSliceFlag{
+		Name:     searchAttributeFlagName,
+		Usage:    fmt.Sprintf("Flag can be used multiple times; value must be \"name=type\"; valid types are: %v", getSearchAttributeTypes()),
+		Aliases:  []string{"sa"},
+		Required: true,
```
mastermanu, line 66: why is this required?

mastermanu: (or where is this getting used? maybe I have missed it)

app/namespace.go
```diff
@@ -145,6 +186,33 @@ func (c *NamespaceClient) parseExistingCerts(ctx *cli.Context) (namespace *names
 	return n, existingCerts, nil
 }
 
+func (c *NamespaceClient) toUserNamespacePermissions(userPermissionsInput map[string]string) ([]*auth.UserNamespacePermissions, error) {
+	var res []*auth.UserNamespacePermissions
+	for email, actionGroup := range userPermissionsInput {
+		u, err := c.authClient.GetUser(c.ctx, &authservice.GetUserRequest{
```
mastermanu, line 192: is this potentially going to cause any rate limiting by the API?

app/namespace.go
```diff
@@ -145,6 +186,33 @@ func (c *NamespaceClient) parseExistingCerts(ctx *cli.Context) (namespace *names
...
+			UserEmail: email,
+		})
+		if err != nil {
+			return nil, err
+		}
+		if len(u.GetUser().GetId()) == 0 {
+			return nil, fmt.Errorf("user not found for: %s", email)
```
mastermanu, line 199: [Nit] consider using multierror so the user can get a list of all the issues that are wrong instead of having to do one at a time

app/namespace.go
```diff
@@ -184,6 +272,125 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+						}
+						if len(certFilterBytes) > 0 {
+							newFilters, err := parseCertificateFilters(certFilterBytes)
+							if err != nil {
+								return err
+							}
+							n.Spec.CertificateFilters = append(n.Spec.CertificateFilters, newFilters.toSpec()...)
+						}
+
+						// search attribute (optional)
```
mastermanu, line 372: [Nit] //search attributes

app/namespace.go
mattkim: this is used by the update namespace command where the flag is required.

app/namespace.go
```diff
@@ -184,6 +272,129 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
 			Subcommands: []*cli.Command{
+				{
+					Name:    "create",
+					Usage:   "Create a temporal namespace",
+					Aliases: []string{"c"},
+					Flags: []cli.Flag{
+						RequestIDFlag,
+						namespaceRegionFlag,
+						namespaceEnvironmentFlag,
+						&cli.StringFlag{
```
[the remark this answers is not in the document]
mattkim: wanted to set a different default value--open question on if we should default to 30?

app/namespace.go
mattkim: will reuse this one.

app/namespace.go
mattkim: removed the flag--you are right it seems to work without it

app/namespace.go
mattkim: it could but what options do we have here? some kind of artificial sleep?

mattkim: agreed offline to move this fetching of users to the server side as part of another pr.

app/namespace.go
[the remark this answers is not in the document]
mattkim: Actually it should be 30 days--reverting this back.

mastermanu (approved):

anekkanti (approved):

PR #155, opened 2023-03-15
Add CreateNamespace command

### LINKED ISSUE #154: [Feature Request] Add CreateNamespace command
### Is your feature request related to a problem? Please describe.

I want to be able to have a bash script (in lieu of having terraform) ensure that the correct namespaces exist in our temporal cloud account. 

We are creating certs and storing them in SSM params, so I want to be able to iterate through the params I can find and either update the certs in the namespaces, or create new namespaces if the one specified doesn't exist.

### Describe the solution you'd like

I want to be able to call it like so:

```sh
aws ssm get-parameter \
  --with-decryption \
  --recursive \
  --name /temporalcloud/my-new-namespace/certificate \
  --no-cli-pager \
  --query "Parameter.Value" \
  --output text \
  > my-new-namespace.crt

temporal namespace create \
  --namespace my-new-namespace \
  --region us-east-1 \
  --retention-days 14 \
  --ca-certificate-file my-new-namespace.crt
```

### Additional context

I am happy to implement or contribute this. I actually have a version that works on my fork, and I'll submit a draft PR after this issue. 

If a create command is not in your design spec, feel free to close this issue & PR. I can just use my forked version as needed (although it's hacky...).

jskrzypek:
I want to be able to programmatically ensure that specific namespaces exist using tcld in our IaC repo based on parameters in AWS SSM, so this adds a `tcld namespace create` command that expects the cert, retention-days, and region flags (I just added the region one).

I was neither able to get the tests that I added to pass, nor to grok why they fail, they look fine to me on visual inspection...

## What was changed
1.  A `RegionFlag` that accepts the target region to use for the new namespace
2. A new private `createNamespace` method on the `NamespaceClient`
3. A subcommand under `NamespaceCommand` that calls NamespaceClient.createNamespace and accepts the `NamespaceFlag`, `RegionFlag`, `RetentionDaysFlag`, `RequestIDFlag`, `CaCertificateFlag`, and `CaCertificateFileFlag` flags
4. A test function that does not work but looks like it ought to, maybe someone can point out where I went wrong?

## Why?
The internal machinery for having a `tcld namespace create` machinery was all hooked up in the protogen code, it's just the final methods in the namespace module that needed to be added.

I did not add flags for all of the members that can be included in a `CreateNamespaceRequest` or `Namespace.Spec`, as we aren't using them, but they could easily be added.

I added tests, but they don't pass and I couldn't understand why, so I left them in in case others can.

## Checklist

1. Closes #154

5. How was this tested:
I tested this first by running this (without building tcld) 
```sh
# I'm sanitizing sensitive bits of these commands
$ aws ssm get-parameter \
  --with-decryption \
  --recursive \
  --name /temporalcloud/namespace01/certificate \
  --no-cli-pager \
  --query "Parameter.Value" \
  --output text \
  > my-new-namespace01.crt

$ temporal namespace create \
  --namespace namespace01 \
  --region us-west-1 \
  --retention-days 14 \
  --ca-certificate-file namespace01.crt
{
        "requestStatus": {
                "requestId": "00000000-0000-0000-0000-000000000000",
                "state": "InProgress",
                "checkDuration": "10s",
                "operationType": "CreateNamespace",
                "resourceId": "namespace01.123ab",
                "resourceType": "namespace",
                "failureReason": "",
                "startTime": "2023-03-15T20:38:51Z",
                "finishTime": null
        }
}
```

Then I checked the UI and it had worked. I then commented out my broken tests and ran `make` to get a binary I could use to script what I needed to.

7. Any docs updates needed?

The [`tcld namespace command reference`](https://docs.temporal.io/cloud/tcld/namespace) page would need an entry for the `create` subcommand.

### DIFF
--- a/app/namespace_test.go
+++ b/app/namespace_test.go
@@ -127,6 +127,101 @@ func (s *NamespaceTestSuite) TestList() {
 	s.NoError(s.RunCmd("namespace", "list"))
 }
 
+func (s *NamespaceTestSuite) TestCreate() {
+
+	type morphCreateReq func(*namespaceservice.CreateNamespaceRequest)
+
+	path := "cafile"
+	s.NoError(os.WriteFile(path, []byte("cert2"), 0644))
+	defer os.Remove(path)
+
+	tests := []struct {
+		name         string
+		args         []string
+		expectErr    bool
+		expectCreate morphCreateReq
+	}{{
+		name:      "displays help",
+		args:      []string{"namespace", "create"},
+		expectErr: true,
+	}, {
+		name:      "missing flags ca, rg, rd",
+		args:      []string{"namespace", "create", "set", "--namespace", "ns1"},
+		expectErr: true,
+	}, {
+		name:      "missing flags rg, rd",
+		args:      []string{"n", "c", "-n", "ns1", "--ca-certificate", "cert1"},
+		expectErr: true,
+	}, {
+		name:      "missing flags ca, rd",
+		args:      []string{"n", "c", "-n", "ns1", "--region", "reg1"},
+		expectErr: true,
+	}, {
+		name:      "missing flag rd",
+		args:      []string{"n", "c", "-n", "ns1", "--ca-certificate", "cert1", "--region", "reg1"},
+		expectErr: true,
+	}, {
+		name:         "create ns1 reg1 rd7 cert1",
+		args:         []string{"namespace", "create", "--namespace", "ns1", "--ca-certificate", "cert1", "--region", "reg1", "--retention-days", "7"},
+		expectCreate: func(r *namespaceservice.CreateNamespaceRequest) {},
+	}, {
+		name: "create ns1 reg2 rd14 cert2",
+		args: []string{"n", "c", "-n", "ns1", "--ca-certificate", "cert2", "--region", "reg2", "--retention-days", "14"},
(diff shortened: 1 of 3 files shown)


jskrzypek:
Closing because this was released in [v0.5.0](https://github.com/temporalio/tcld/releases/tag/v0.5.0)

app/namespace_test.go
```diff
@@ -127,6 +127,101 @@ func (s *NamespaceTestSuite) TestList() {
...
+				req := namespaceservice.CreateNamespaceRequest{
+					Namespace: "ns1",
+					Spec: &namespace.NamespaceSpec{
+						AcceptedClientCa: "cert1",
+						Region:           "reg1",
+						RetentionDays:    int32(7),
+					},
+				}
+				tc.expectCreate(&req)
+				s.mockService.EXPECT().CreateNamespace(gomock.Any(), &req).
```
mattkim, line 209: I believe the issue with the unit test is because `&req` is a pointer passed in here but it will never match the pointer being passed in at runtime. You would need to do a deep equal check using `gomock.Matcher`.

jskrzypek:
Ahhh ok that makes sense. In short: I'd hit the limits of my knowledge of
how to write tests in Golang 😛

PR #158, opened 2023-03-23
add namespace delete command

mattkim:
## What was changed

add namespace delete command

## Why?

allow users to delete a namespace using `tcld`

## Checklist
1. How was this tested:

unit tests


app/namespace.go
```diff
@@ -184,6 +200,29 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
 			},
 			Subcommands: []*cli.Command{
+				{
+					Name:    "delete",
+					Usage:   "Delete a temporal namespace",
+					Aliases: []string{"d"},
+					Flags: []cli.Flag{
+						RequestIDFlag,
+						ResourceVersionFlag,
+						&cli.StringFlag{
```
mastermanu, line 210: Should we ask the user to confirm deletion by typing in namespace name (and then I believe we have an auto-confirm flag if user wants to bypass it)?

mastermanu (approved):

anekkanti (approved):

PR #162, opened 2023-03-29
add user management commands

mattkim:
## What was changed

add user management commands to `tcld`

Commands added:
- user
  - list: list all users in the account
  - get: get a user in the account
  - invite: invite users to the account along with account role and namespace permissions
  - delete: delete a user from the account
  - resend-invite: resend invite for a user
  - roles
    - list: list all roles for a user
    - set-account-role: set the account role for a user
    - set-namespace-permissions: set namespace permissions for a user

## Why?

give users the ability to automate user management with tcld

## Checklist
1. How was this tested:
  unit tests


app/user.go
```diff
@@ -0,0 +1,597 @@
...
+		RequestId: ctx.String(RequestIDFlagName),
+	}
+	resp, err := c.client.UpdateUserNamespacePermissions(c.ctx, req)
+	if err != nil {
+		return fmt.Errorf("unable to update user's namespace permission: %w", err)
+	}
+	return PrintProto(resp.GetRequestStatus())
+}
+
+func (c *UserClient) setNamespacePermissions(
```
mattkim, line 323: this doesn't seem like it's actually doing a set :P

mattkim: i wonder if we should have an `update` and a `set` api.

app/role.go
```diff
@@ -0,0 +1,128 @@
...
+	return res, nil
+}
+
+func getNamespaceRole(ctx context.Context, client authservice.AuthServiceClient, namespace string, actionGroup string) (*auth.Role, error) {
+	ag, err := toNamespaceActionGroup(actionGroup)
+	if err != nil {
+		return nil, err
+	}
+	res, err := client.GetRolesByPermissions(ctx, &authservice.GetRolesByPermissionsRequest{
+		Specs: []*auth.RoleSpec{{
```
mastermanu, line 55: can we not do batching and invoke the client less often? Or is that not worth it?

app/role.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return rv
+}
+
+func getNamespaceRolesFromMap(ctx context.Context, client authservice.AuthServiceClient, namespaceActionGroups map[string]string) ([]*auth.Role, error) {
+	var res []*auth.Role
+	for namespace, actionGroup := range namespaceActionGroups {
+		r, err := getNamespaceRole(ctx, client, namespace, actionGroup)
+		if err != nil {
+			return nil, err
```
mastermanu, line 42: [Nit] is a multierr a better user experience here?

app/role.go
```diff
@@ -0,0 +1,128 @@
...
+		}},
+	})
+	if err != nil {
+		return nil, err
+	}
+	if len(res.Roles) == 0 {
+		return nil, fmt.Errorf("no roles found")
+	}
+	if len(res.Roles) > 1 {
+		return nil, fmt.Errorf("more than 1 account role found: %s", res.Roles)
```
mastermanu, line 69: [Nit] typo in error message?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+		}
+		for _, nsRole := range nsRoles {
+			roleIDs = append(roleIDs, nsRole.GetId())
+		}
+	}
+
+	// Invite users to given roles
+	req := &authservice.InviteUsersRequest{
+		Specs:     make([]*auth.UserSpec, len(emails)),
+		RequestId: ctx.String(RequestIDFlagName),
```
tminusplus, line 160: Should `RequestIDFlagName` be required, since we directly use it in most of these calls?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+	res, err := c.client.GetUser(c.ctx, &authservice.GetUserRequest{
+		UserEmail: userEmail,
+		UserId:    userID,
+	})
+	if err != nil {
+		return err
+	}
+	req := &authservice.DeleteUserRequest{
+		UserId:          res.User.Id,
+		ResourceVersion: ctx.String(ResourceVersionFlagName),
```
tminusplus, line 216: Should `ResourceVersionFlagName` be required since we use it directly in this command?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+	return PrintProto(resp.GetRequestStatus())
+}
+
+func (c *UserClient) setAccountRole(
+	ctx *cli.Context,
+	userID string,
+	userEmail string,
+	accountRole string,
+) error {
+	user, userRoles, err := c.getUserAndRoles(userID, userEmail)
```
tminusplus, line 252: Should we make sure one of these is set (or use that MutuallyExclusiveFlag + required that I mentioned earlier)?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+			}
+			// find the current admin role to replace
+			ar := ur.Spec.AccountRole
+			if ar != nil && ar.ActionGroup != auth.ACCOUNT_ACTION_GROUP_UNSPECIFIED {
+				// only swap the current admin role
+				userRoles[i] = accountRoleToSet
+			}
+		}
+	}
+	roleNames := make([]string, len(userRoles))
```
tminusplus, line 287: nit: might be nicer to allocate capacity `make([]string, 0, len(userRoles))` then append to avoid.

app/user.go
tminusplus: I'm more in favor of calling this `update` and allow the behavior to be decided in the API (upsert or not). I like the AIP guidelines about this, which follows the declarative model we have. https://google.aip.dev/134#create-or-update

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+					},
+				},
+				{
+					Name:    "invite",
+					Usage:   "Invite users to Temporal Cloud",
+					Aliases: []string{"i"},
+					Flags: []cli.Flag{
+						&cli.StringSliceFlag{
+							Name:     userEmailFlagName,
+							Usage:    "The email address of the user, you can supply this flag multiple times to invite multiple users in a single request",
```
tminusplus, line 430: nit: `StringSliceFlag` mentions that it accepts multiple inputs so I think we could shorten this to `The email address(es) of the user to invite`

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+	userEmailFlagName           = "user-email"
+	accountRoleFlagName         = "account-role"
+	namespacePermissionFlagName = "namespace-permission"
+)
+
+var (
+	userIDFlag = &cli.StringFlag{
+		Name:    userIDFlagName,
+		Usage:   "The user id",
+		Aliases: []string{"i"},
```
mastermanu, line 25: [Nit] I think 'id' is better than 'i'

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+						},
+						&cli.StringFlag{
+							Name:     accountRoleFlagName,
+							Usage:    fmt.Sprintf("The account role to set on the user; valid types are: %v", accountActionGroups),
+							Aliases:  []string{"ar"},
+							Required: true,
+						},
+						&cli.StringSliceFlag{
+							Name:    namespacePermissionFlagName,
+							Usage:   fmt.Sprintf("Flag can be used multiple times; value must be \"namespace=permission\"; valid types are: %v", namespaceActionGroups),
```
tminusplus, line 442: Could we use a [StringMapFlag](https://pkg.go.dev/github.com/urfave/cli/v3#StringMapFlag) instead? It will enforce the format `namespace=permission`.

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+	"github.com/temporalio/tcld/protogen/api/authservice/v1"
+	"github.com/urfave/cli/v2"
+	"strings"
+)
+
+const (
+	userIDFlagName              = "user-id"
+	userEmailFlagName           = "user-email"
+	accountRoleFlagName         = "account-role"
+	namespacePermissionFlagName = "namespace-permission"
```
mastermanu, line 18: I think we added a --user-namespace-permission flag for namespaces? Do we want to try and consolidate?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+
+var (
+	userIDFlag = &cli.StringFlag{
+		Name:    userIDFlagName,
+		Usage:   "The user id",
+		Aliases: []string{"i"},
+	}
+	userEmailFlag = &cli.StringFlag{
+		Name:    userEmailFlagName,
+		Usage:   "The user email address of the user to be invited",
```
mastermanu, line 29: is the usage of this flag guaranteed to ALWAYS be in the context of invitations?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+							Usage:   "Delete namespace permissions for a user",
+							Aliases: []string{"dnp"},
+							Flags: []cli.Flag{
+								userIDFlag,
+								userEmailFlag,
+								RequestIDFlag,
+								ResourceVersionFlag,
+								&cli.StringSliceFlag{
+									Name:     NamespaceFlagName,
+									Usage:    "The namespace to delete user permissions from",
```
tminusplus, line 578: nit:  `The namespace to delete` -> `The namespace(s) to delete`

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+							Aliases:  []string{"e"},
+							Required: true,
+						},
+						&cli.StringFlag{
+							Name:     accountRoleFlagName,
+							Usage:    fmt.Sprintf("The account role to set on the user; valid types are: %v", accountActionGroups),
+							Aliases:  []string{"ar"},
+							Required: true,
+						},
+						&cli.StringSliceFlag{
```
tminusplus, line 440: Could be nice to define this once since we use it in multiple places.

app/user.go
tminusplus, line 442: nit: the help already mentions it accepts multiple inputs, could be nice to describe what it does `sets the user's permission for the namespace`.

app/role.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return res.Roles[0], nil
+}
+
+func toAccountActionGroup(actionGroup string) (auth.AccountActionGroup, error) {
```
tminusplus, line 98: nit: I think on all the action group flags, we could add a `SuggestFlagFunc` for them which provides the accepted values. Should make validation easier and I think this enables auto-complete in user's terminals when using our cli.

tminusplus (approved):
Mostly style / small comments. I'm happy to come back and address some of the comments later if we need to get this in sooner.

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+		Aliases: []string{"i"},
+	}
+	userEmailFlag = &cli.StringFlag{
+		Name:    userEmailFlagName,
+		Usage:   "The user email address of the user to be invited",
+		Aliases: []string{"e"},
+	}
+)
+
+type UserClient struct {
```
mastermanu, line 34: do we follow this same pattern for other resource types? Also, why is this exported

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+		}
+	}
+}
+
+func (c *UserClient) getUser(userID, userEmail string) (*auth.User, error) {
+	if userID == "" && userEmail == "" {
+		return nil, fmt.Errorf("both user-id and user-email not set")
+	}
+
+	if userID != "" && userEmail != "" {
```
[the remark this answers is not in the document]
mastermanu: this is pretty neat, but we are currently using v2, and it seems the v3 package is alpha. Does this exist in v2?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+		pageToken = res.NextPageToken
+		if len(pageToken) == 0 {
+			return PrintProto(totalRes)
+		}
+	}
+}
+
+func (c *UserClient) getUser(userID, userEmail string) (*auth.User, error) {
+	if userID == "" && userEmail == "" {
+		return nil, fmt.Errorf("both user-id and user-email not set")
```
mastermanu, line 76: [Nit] error message implies that user should set both. we should say exactly one must be set

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+		return errors.New("account role required for inviting new users")
+	}
+
+	// the role ids to invite the users for
+	var roleIDs []string
+
+	// first get the required account role
+	role, err := getAccountRole(c.ctx, c.client, accountRole)
+	if err != nil {
+		return err
```
mastermanu, line 138: is the error returned from this clear to someone trying to invite a user? Should we wrap the error message here with "invalid account role specified: %v" ?

app/user.go
mastermanu: the server assigns one automatically if its not supplied by creating a new UUID. What we could do if its not specified is generate the UUID ourselves maybe?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+	return PrintProto(resp.GetRequestStatus())
+}
+
+func (c *UserClient) deleteUser(
+	ctx *cli.Context,
+	userID string,
+	userEmail string,
+) error {
+	if userID != "" && userEmail != "" {
```
mastermanu, line 200: can we just call getUser() to avoid duplicating this code?

app/user.go
mastermanu: its optional given we populate it from the existing user if the user doesn't supply a specific version

app/user.go
mastermanu: getUserAndRoles is doing that check underneath the covers, right?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+		if userRoles[i].Id == accountRoleToSet.Id {
+			return fmt.Errorf("user already has '%s' account role", accountRoleToSet.Id)
+		}
+	}
+	// check if this is the global admin role, and replace all existing roles
+	if accountRoleToSet.Spec.AccountRole.ActionGroup == auth.ACCOUNT_ACTION_GROUP_ADMIN {
+		// set the user account admin role
+		y, err := ConfirmPrompt(ctx, "Setting admin role on user, please confirm")
+		if err != nil || !y {
+			return err
```
mastermanu, line 270: if err == nil and y = false, we are going to return a nil error? That's probably fine, but then maybe we should print out "operation canceled" or something like that?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+	for i := range userRoles {
+		if userRoles[i].Id == accountRoleToSet.Id {
+			return fmt.Errorf("user already has '%s' account role", accountRoleToSet.Id)
+		}
+	}
+	// check if this is the global admin role, and replace all existing roles
+	if accountRoleToSet.Spec.AccountRole.ActionGroup == auth.ACCOUNT_ACTION_GROUP_ADMIN {
+		// set the user account admin role
+		y, err := ConfirmPrompt(ctx, "Setting admin role on user, please confirm")
```
mastermanu, line 268: should the message should indicate all existing namespace permissions will be replaced?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+		if err != nil || !y {
+			return err
+		}
+		userRoles = []*auth.Role{accountRoleToSet}
+	} else {
+		for i := range userRoles {
+			ur := userRoles[i]
+			if ur.Type != auth.ROLE_TYPE_PREDEFINED {
```
mastermanu, line 276: can you remind me what this check is for?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+				continue
+			}
+			// find the current admin role to replace
+			ar := ur.Spec.AccountRole
+			if ar != nil && ar.ActionGroup != auth.ACCOUNT_ACTION_GROUP_UNSPECIFIED {
+				// only swap the current admin role
+				userRoles[i] = accountRoleToSet
```
mastermanu, line 283: should we break after this assignment?
also is there any scenario where we exit the for loop without doing any update at all? should we error if that happens?

mastermanu: also whats the 2nd check for? in case there is a future admin role we add that the CLI doesn't understand? But in that case, then CLI cannot be used to update admin role since it can never replace it?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+	userEmail string,
+	accountRole string,
+) error {
+	user, userRoles, err := c.getUserAndRoles(userID, userEmail)
+	if err != nil {
+		return err
+	}
+	accountRoleToSet, err := getAccountRole(c.ctx, c.client, accountRole)
+	if err != nil {
+		return err
```
mastermanu, line 258: same question, it's not obvious if an error returned from the underlying call has enough context to be friendly to the user

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+	npm, err := toNamespacePermissionsMap(namespacePermissions)
+	if err != nil {
+		return err
+	}
+	for namespace, permission := range npm {
+		ag, err := toNamespaceActionGroup(permission)
+		if err != nil {
+			return err
+		}
+		if err := c.updateUserNamespacePermissions(ctx, userID, userEmail, namespace, ag); err != nil {
```
mastermanu, line 338: isn't this another case where we can do a batch update here?

mastermanu: actually, I am pretty sure we HAVE to do this in a single batch

mastermanu: nevermind, the above statement is wrong. That API can actually be used btw to update multiple permissions either for the same user or across users

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+}
+
+func (c *UserClient) deleteNamespacePermission(
+	ctx *cli.Context,
+	userID string,
+	userEmail string,
+	namespaces []string,
+) error {
+	for _, ns := range namespaces {
+		if err := c.updateUserNamespacePermissions(ctx, userID, userEmail, ns, auth.NAMESPACE_ACTION_GROUP_UNSPECIFIED); err != nil {
```
mastermanu, line 352: wait, our way to update user permissions is to set the action group to unspecified?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+		Namespace: namespace,
+		UserNamespacePermissions: []*auth.UserNamespacePermissions{
+			{
+				UserId:      user.Id,
+				ActionGroup: actionGroup,
+			},
+		},
+		RequestId: ctx.String(RequestIDFlagName),
+	}
+	resp, err := c.client.UpdateUserNamespacePermissions(c.ctx, req)
```
mastermanu, line 316: does this also check if the user already has the existing permissions? it seems server will fail, so its not really an upsert?

app/user.go
```diff
@@ -0,0 +1,597 @@
...
+	for _, kv := range keyValues {
+		parts := strings.Split(kv, "=")
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid namespace permission \"%s\" must be of format: \"namespace=permission\"", kv)
+		}
+
+		namespace := parts[0]
+		actionGroupValue := parts[1]
+
+		if len(namespace) == 0 {
```
mastermanu, line 370: do we need to check parts[1] length?

mastermanu (changes requested):
left some comments

app/print.go
```diff
@@ -25,15 +26,47 @@ func PrintObj(i interface{}) error {
...
+func PrintProto(m proto.Message) error {
+	ser, err := serializeProto(m)
 	if err != nil {
 		return err
 	}
 	fmt.Printf("%v\n", ser)
 	return nil
 }
+
+func PrintProtoSlice(name string, ms []proto.Message) error {
```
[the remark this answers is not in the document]
mattkim: Yeah because the slice doesn't support the proto.Message interface.

Interesting not about `gogo/protobuf`--yes we should deprecate it.

app/role.go
mattkim: changed to use batch api so not relevant now.

app/role.go
mattkim: So this is very intriguing--but i couldn't figure out how to get it working and don't see any documentation. Do you have any references?

app/user.go
mattkim: unfortunately not reusable because one flag is required the other it is not.

app/user.go
mattkim: interesting--not available in v2--but something we should look into supporting.

app/user.go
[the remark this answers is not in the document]
tminusplus: Doh you are right. Honestly we have been finding some issues with urfave (if you have an env variable which stores a secret it will print that secret in the help message)... but that is a separate discussion from this PR.

app/user.go
tminusplus: Server assigning one sounds best, that makes sense!

app/user.go
tminusplus: Also doh, searching for urfave go docs brought me to v3.

app/role.go
tminusplus: You know what I tried to look for this when me and @anekkanti discovered that urfave prints env variables into the help message.. and couldn't find anything on it.

app/version.go
```diff
@@ -3,7 +3,7 @@ package app
 import "github.com/urfave/cli/v2"
 
 const (
-	DefaultVersion = "v0.5.0"
+	DefaultVersion = "v0.6.0"
```
tminusplus, line 6: I do wonder if we should cut releases (bump the version number) in a separate PR. It helps keep the commit history clean and easy to understand at a glance.

tminusplus (approved):
Minor comments otherwise LGTM

app/user.go
```diff
@@ -0,0 +1,560 @@
...
+	"github.com/urfave/cli/v2"
+	"strings"
+)
+
+const (
+	userIDFlagName              = "user-id"
+	userEmailFlagName           = "user-email"
+	accountRoleFlagName         = "account-role"
+	namespacePermissionFlagName = "namespace-permission"
+	permissionFlagName          = "permission"
```
tminusplus, line 18: Is this flag used anywhere? I tried looking and it seems unused.

mastermanu (approved):

app/user.go
mattkim: Removed.

PR #168, opened 2023-04-17
add support for apikeys

anekkanti:
```
$ ./tcld user apikey create
NAME:
   tcld user apikey create - Create a user apikey

USAGE:
   tcld user apikey create [command options] [arguments...]

OPTIONS:
   --name value, -n value             the display name of the apikey
   --description value, --desc value  the description of the apikey
   --duration value, -d value         the duration from now when the apikey will expire, will be ignored if expiry flag is set (default: 0s)
   --expiry value, -e value           the absolute timestamp (RFC3339) when the apikey will expire, example: '2023-04-26T12:41:27-07:00'
   --request-id value, -r value       The request-id to use for the asynchronous operation, if not set the server will assign one (optional)
   --help, -h                         show help
   ```

```
$ ./tcld user apikey delete
NAME:
   tcld user apikey delete - Delete an user apikey

USAGE:
   tcld user apikey delete [command options] [arguments...]

OPTIONS:
   --id value, -i value                The id of the apikey to delete
   --resource-version value, -v value  The resource-version (etag) to update from, if not set the cli will use the latest (optional)
   --request-id value, -r value        The request-id to use for the asynchronous operation, if not set the server will assign one (optional)
   --help, -h                          show help
```

```
$ ./tcld user apikey list
NAME:
   tcld user apikey list - List user apikeys

USAGE:
   tcld user apikey list [command options] [arguments...]

OPTIONS:
   --help, -h  show help
```

```
$ ./tcld user apikey get
NAME:
   tcld user apikey get - Get an user apikey

USAGE:
   tcld user apikey get [command options] [arguments...]

OPTIONS:
   --id value, -i value  The id of the apikey to get
   --help, -h            show help
```

### CODE REVIEW

File: app/apikey.go [hunk 1]
```diff
@@ -0,0 +1,231 @@
...
+			Usage:   "APIKey operations",
+			Before: func(ctx *cli.Context) error {
+				var err error
+				c, err = getAPIKeyClientFn(ctx)
+				return err
+			},
+			Subcommands: []*cli.Command{
+				{
+					Name:    "create",
+					Usage:   "Create an user apikey",
```
mastermanu, line 124: [Nit] "a user"

mastermanu: https://www.scribbr.com/commonly-confused-words/a-vs-an/#:~:text=A%20is%20used%20before%20a%20noun%20that%20starts,with%20a%20vowel%20sound%20%28e.g.%2C%20%E2%80%9Ca%2C%E2%80%9D%20%E2%80%9Co%2C%E2%80%9D%20%E2%80%9Ci%E2%80%9D%29.

app/apikey.go
```diff
@@ -0,0 +1,231 @@
...
+						RequestIDFlag,
+					},
+					Action: func(ctx *cli.Context) error {
+						expiry := ctx.Timestamp("expiry")
+						if expiry == nil || expiry.IsZero() {
+							expiryPeriod := ctx.Duration("duration")
+							if expiryPeriod == 0 {
+								return fmt.Errorf("no expiry was set")
+							}
+							e := time.Now().Add(expiryPeriod)
```
mastermanu, line 158: UTC?

app/apikey.go
```diff
@@ -0,0 +1,231 @@
...
+							Aliases: []string{"desc"},
+						},
+						&cli.DurationFlag{
+							Name:    "duration",
+							Usage:   "the duration from now when the apikey will expire, will be ignored if expiry flag is set",
+							Aliases: []string{"d"},
+						},
+						&cli.TimestampFlag{
+							Name:    "expiry",
+							Usage:   "the absolute timestamp when the apikey will expire, example: 2023-04-12T15:14:05",
```
mastermanu, line 145: [Nit] is this considered UTC time by default?

mastermanu (dismissed):

app/apikey.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (s *APIKeyClient) createAPIKey(
+	displayName string,
+	description string,
+	expiry time.Time,
+	requestID string,
+) error {
+	expiryts, err := types.TimestampProto(expiry)
+	if err != nil {
+		return fmt.Errorf("failed to convert timestamp to proto: %s", err)
```
tminusplus, line 45: Could be good to wrap the error `%w`

app/apikey.go
```diff
@@ -0,0 +1,231 @@
...
+			},
+			Subcommands: []*cli.Command{
+				{
+					Name:    "create",
+					Usage:   "Create an user apikey",
+					Aliases: []string{"c"},
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:     "name",
+							Usage:    "the display name of the apikey",
```
tminusplus, line 129: nit: I wonder if we should replace all these instances of `apikey` with `API key` or some other form and keep it consistent in the other help messages we have in the CLI.

app/apikey.go
```diff
@@ -0,0 +1,231 @@
...
+		if len(pageToken) == 0 {
+			return PrintProto(totalRes)
+		}
+	}
+}
+
+func (s *APIKeyClient) getAPIKey(
+	id string,
+) (*auth.APIKey, error) {
+	resp, err := s.authClient.GetAPIKey(s.ctx, &authservice.GetAPIKeyRequest{
```
tminusplus, line 84: Could be good to check that the ID is a non-zero length.

app/apikey.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	return resp.ApiKey, nil
+}
+
+func (s *APIKeyClient) deleteAPIKey(
+	id string,
+	resourceVersion string,
+	requestID string,
+) error {
+	resp, err := s.authClient.DeleteAPIKey(s.ctx, &authservice.DeleteAPIKeyRequest{
```
tminusplus, line 98: Ditto on length checks for ID here

app/apikey.go
```diff
@@ -0,0 +1,231 @@
...
+	expiry time.Time,
+	requestID string,
+) error {
+	expiryts, err := types.TimestampProto(expiry)
+	if err != nil {
+		return fmt.Errorf("failed to convert timestamp to proto: %s", err)
+	}
+	resp, err := s.authClient.CreateAPIKey(s.ctx, &authservice.CreateAPIKeyRequest{
+		Spec: &auth.APIKeySpec{
+			DisplayName: displayName,
```
tminusplus, line 49: I think in the UX design we said the display name is required, could be good to do a quick length check here

go.mod
```diff
@@ -7,14 +7,14 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/kylelemons/godebug v1.1.0
 	github.com/stretchr/testify v1.7.1
-	github.com/urfave/cli/v2 v2.8.1
+	github.com/urfave/cli/v2 v2.25.1
```
tminusplus, line 10: Oh interesting I wonder where this downgrade came from

app/apikey.go
anekkanti: I think the user needs to explicitly specify the time zone in the input, we should be good.

app/apikey.go
anekkanti: The ID is set as `required`. urfave/cli package will make sure its never empty. Not point of adding a redundant check.

app/apikey.go
anekkanti: The displayname is set as `required`. urfave/cli package will make sure its never empty. Not point of adding a redundant check.

go.mod
anekkanti: was probably some bad merge.

tminusplus (approved):

PR #169, opened 2023-04-17
Add bearer token API keys and move authentication credentials into PerRPCCredentials

tminusplus:
## What was changed
This PR achieves several things:
1. Adds support for bearer token API keys in the format of`$prefix_$id_$secret`.
2. Moves all authentication credentials into PerRPCCredentials to help prevent accidental leakage of secrets in code and to enforce secure connections when they are used.
3. Refactors the `GetServerConnection` tests from running a `urfave app` to setting up a `cli.Context` which parses flags and then connects to a server via `bufconn`.
4. Adds an `--insecure` flag, hidden to discourage use, to allow sending authentication over insecure connections. This is important for users who are using proxies like envoy which then sets up a secure connection from `localhost`.

## Why?

To add support, and tests, for our beta API keys in a way which encourages secure usage by default.

## Checklist
N/A

## Test Plan

- [x] Added tests for `oauth`, `apikeys`, and `HMAC`.
- [x] Tested a tcld command with `oauth` creds.
- [x] Tested a tcld command with `apikey` creds.

Since HMAC / request signatures was punted, I did not test the functionality but instead moved it over verbatim into the `apikey` credential to keep the refactor from being destructive.


app/connection.go
```diff
@@ -4,67 +4,93 @@ import (
...
+	return ctx, conn, nil
+}
+
+func defaultDialOptions(c *cli.Context, addr *url.URL) ([]grpc.DialOption, error) {
+	var opts []grpc.DialOption
+
+	creds, err := newRPCCredential(c)
 	if err != nil {
-		return nil, nil, err
+		return []grpc.DialOption{}, nil
```
mastermanu, line 53: is this going to be confusing behavior to the user? Why did we deviate from returning an error?

app/credentials/apikey/apikey.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if !c.insecure {
+		// Ensure the API key, AKA bearer token, is sent over a secure connection - meaning TLS.
+		if err := credentials.CheckSecurityLevel(ri.AuthInfo, credentials.PrivacyAndIntegrity); err != nil {
+			return nil, fmt.Errorf("the connection's transport security level is too low for API keys: %v", err)
+		}
+	}
+
+	if c.enableHMAC {
```
mastermanu, line 90: this feels premature to check-in to the code base?

app/connection.go
tminusplus: Ah that is a mistake, it should be returning the error. Will fix.

app/credentials/apikey/apikey.go
```diff
@@ -0,0 +1,140 @@
...
+	// RequestDatetime
+	// ActionName
+	requestSignatureFormat = "%s\n%s\n%s\n%s"
+)
+
+type Credential struct {
+	ID         string
+	secret     string // secret kept private to prevent accidental access.
+	enableHMAC bool
+	insecure   bool
```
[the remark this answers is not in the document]
tminusplus: Will do!

app/credentials/apikey/apikey.go
tminusplus: It was previously in the codebase, so I wanted to move it along to the PerRPCCredentials so this PR wasn't destructive. Then my thought was we can remove it in a future PR when desired.

app/credentials/oauth/oauth.go
```diff
@@ -0,0 +1,68 @@
...
+
+func WithInsecure(insecure bool) Option {
+	return func(c *Credential) {
+		c.insecure = insecure
+	}
+}
+
+type Credential struct {
+	accessToken string // keep unexported to prevent accidental leakage of the token.
+	insecure    bool
```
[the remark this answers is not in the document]
tminusplus: Will do, I initially did it to follow `curl`'s standard arg for this and have it follow through the code but `allowInsecureTransport` is nicer.

mastermanu (dismissed):

anekkanti (approved):

PR #184, opened 2023-05-04
Refactor to single token, use authorization header, and remove request signatures

tminusplus:
## What was changed

This refactors to single token, uses an authorization header over custom headers, and removes the request signatures POC.

## Why?

Prepare for the preview release of API keys.

## Checklist
- [x] Unit tests updated to reflect single token + authorization headers.


app/connection.go
```diff
@@ -70,15 +70,10 @@ func defaultDialOptions(c *cli.Context, addr *url.URL) ([]grpc.DialOption, error
...
-		return nil, fmt.Errorf("when using an API key you must specify both the key ID and secret")
-	} else if len(apiKeyID) > 0 && len(apiSecretKey) > 0 {
+	apiKey := c.String(APIKeyFlagName)
+	if len(apiKey) > 0 {
 		return apikey.NewCredential(
-			apiKeyID,
-			apiSecretKey,
-			apikey.WithHMAC(c.Bool(EnableHMACFlagName)),
+			apiKey,
 			apikey.WithInsecureTransport(insecure),
```
mattkim, line 77: What does this insecure transport do again?

app/credentials/apikey/apikey.go
```diff
@@ -2,64 +2,37 @@ package apikey
...
-	RequestDatetimeHeader            = "tmprl-request-datetime"
-	DefaultRequestSignatureAlgorithm = "tmprl-hmac-sha256"
-
-	// requestSignatureFormat requires each of the following data on a newline
-	// KeyID
-	// RequestSignatureAlgorithm
-	// RequestDatetime
-	// ActionName
-	requestSignatureFormat = "%s\n%s\n%s\n%s"
+	AuthorizationHeader       = "authorization"
```
mattkim, line 11: Does this need to be "A" or does casing not matter

mattkim (dismissed):

app/connection.go
tminusplus: It enables the use of tcld with TLS termination proxies, which some customers may have. Insecure will be set to false by default and the flag is hidden from the user to discourage use.

app/credentials/apikey/apikey.go
tminusplus: Casing should not matter as headers are normalized in gPRC but I will make it capitalized to be consistent.

mattkim (approved):

PR #187, opened 2023-05-09
Commands to generate ca and client certs

anekkanti:
## What was changed
Generate tls certificates using `tcld`

### To create a new ca certificate

```
tcld gen ca --org temporal -d 1y --ca-cert ca.pem --ca-key ca.key
```

### To create a new client certificate
```
tcld gen leaf --org temporal -d 1y --ca-cert ca.pem --ca-key ca.key --cert client.pem --key client.key
```

## TODOs
- [x] Unit tests
- [x] Support elliptic curve certificates as the default
- [x] Security Review


app/certificates.go
```diff
@@ -0,0 +1,429 @@
...
+	"io/ioutil"
+	"math/big"
+	"time"
+
+	"github.com/go-playground/validator/v10"
+	"github.com/temporalio/tcld/utils"
+	"github.com/urfave/cli/v2"
+)
+
+func generateRandomString(n int) (string, error) {
```
mastermanu, line 22: can we be consistent with how we generate random strings with what we are doing in the auth package?

app/certificates.go
```diff
@@ -0,0 +1,429 @@
...
+func generateCACertificate(
+	input generateCACertificateInput,
+) (caPEM, caPrivateKeyPEM []byte, err error) {
+	validator := validator.New()
+	if err := validator.Struct(input); err != nil {
+		return nil, nil, err
+	}
+	randomLetters, err := generateRandomString(4)
+	if err != nil {
+		return nil, nil, err
```
mastermanu, line 51: [Nit] wrap error with more friendly error message for user (e.g. unable to generate random   string for dns name or something like that?)

app/certificates.go
```diff
@@ -0,0 +1,429 @@
...
+	keyUsage := x509.KeyUsageDigitalSignature | x509.KeyUsageCRLSign | x509.KeyUsageCertSign
+	if input.RSAAlgorithm {
+		// Only RSA subject keys should have the KeyEncipherment KeyUsage bits set. In
+		// the context of TLS this KeyUsage is particular to RSA key exchange and
+		// authentication.
+		keyUsage |= x509.KeyUsageKeyEncipherment
+	}
+
+	dnsRoot := fmt.Sprintf("client.root.%s.%s", input.Organization, randomLetters)
+	now := time.Now()
```
mastermanu, line 63: [Nit] .UTC()?

app/certificates.go
```diff
@@ -0,0 +1,429 @@
...
+
+	keyUsage := x509.KeyUsageDigitalSignature | x509.KeyUsageCRLSign | x509.KeyUsageCertSign
+	if input.RSAAlgorithm {
+		// Only RSA subject keys should have the KeyEncipherment KeyUsage bits set. In
+		// the context of TLS this KeyUsage is particular to RSA key exchange and
+		// authentication.
+		keyUsage |= x509.KeyUsageKeyEncipherment
+	}
+
+	dnsRoot := fmt.Sprintf("client.root.%s.%s", input.Organization, randomLetters)
```
mastermanu, line 62: are there any limitations on organization characters?

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+	now := time.Now()
+	conf := &x509.Certificate{
+		SerialNumber: big.NewInt(2019),
```
mastermanu, line 70: How did we get 2019? Doesn't seem to follow spec: https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.2

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+		}
+		ret[i] = letters[num.Int64()]
+	}
+
+	return string(ret), nil
+}
+
+type generateCACertificateInput struct {
+	Organization string        `validate:"required"`
+	Duration     time.Duration `validate:"required"`
```
mastermanu, line 43: I assume we are checking elsewhere that this has a minimum value?

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+	now := time.Now()
+	conf := &x509.Certificate{
+		SerialNumber: big.NewInt(2019),
+		Subject: pkix.Name{
+			Organization: []string{input.Organization},
+		},
+		NotBefore:             now.Add(-time.Minute), // grace of 1 min
+		NotAfter:              now.Add(input.Duration),
+		IsCA:                  true,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
```
mastermanu, line 77: this is a CA certificate, right? Given that this is not going to be used as the client cert, does this ExtKeyUsage need to be set?

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+	if input.RSAAlgorithm {
+		// Only RSA subject keys should have the KeyEncipherment KeyUsage bits set. In
+		// the context of TLS this KeyUsage is particular to RSA key exchange and
+		// authentication.
+		keyUsage |= x509.KeyUsageKeyEncipherment
+	}
+
+	dnsRoot := fmt.Sprintf("client.root.%s.%s", input.Organization, randomLetters)
+	now := time.Now()
+	conf := &x509.Certificate{
```
mastermanu, line 69: are we setting MaxPathLen=1 anywhere?

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+		IsCA:                  true,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		KeyUsage:              keyUsage,
+		BasicConstraintsValid: true,
+		DNSNames:              []string{dnsRoot},
+	}
+
+	var privateKey any
+	if input.RSAAlgorithm {
+		privateKey, err = rsa.GenerateKey(rand.Reader, 4096)
```
mastermanu, line 85: iirc from discussion, we were thinking 2048/256? is that still the case, or odid we change it?

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+	switch k := privateKey.(type) {
+	case *rsa.PrivateKey:
+		publicKey = &k.PublicKey
+	case *ecdsa.PrivateKey:
+		publicKey = &k.PublicKey
+	}
+
+	cert, err := x509.CreateCertificate(rand.Reader, conf, conf, publicKey, privateKey)
+	if err != nil {
+		return nil, nil, err
```
mastermanu, line 103: [Nit] consider wrapping error here as well and other places where it might not be obvious, especially if there is a chance the error is something like EOF or any other cryptic message

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+	return caPEMBuffer.Bytes(), caPrivateKeyPEMBuffer.Bytes(), nil
+}
+
+type generateCertificateInput struct {
+	Organization     string `validate:"required"`
+	OrganizationUnit string
```
mastermanu, line 130: where we going to remove this initially?

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	subject := pkix.Name{
+		Organization:       []string{input.Organization},
+		OrganizationalUnit: []string{input.OrganizationUnit},
+	}
+	dnsRoot := fmt.Sprintf("client.endentity.%s.%s", input.Organization, randomLetters)
+	now := time.Now()
+	conf := &x509.Certificate{
```
mastermanu, line 180: we didnt need extended usage here?

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+		return nil, nil, err
+	}
+
+	privBytes, err := x509.MarshalPKCS8PrivateKey(privateKey)
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to marshal private key: %v", err)
+	}
+	certPrivateKeyPEMBuffer := new(bytes.Buffer)
+	err = pem.Encode(certPrivateKeyPEMBuffer, &pem.Block{
+		Type:  "PRIVATE KEY",
```
mastermanu, line 226: [Nit] can we define these as constants?

app/certificates.go
```diff
@@ -0,0 +1,441 @@
...
+	certPEMBuffer := new(bytes.Buffer)
+	err = pem.Encode(certPEMBuffer, &pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: cert,
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	privBytes, err := x509.MarshalPKCS8PrivateKey(privateKey)
```
mastermanu, line 220: note: iirc Java SDK requires PKCS12 cert format....can you confirm with SDK folks if that is the case? If so, might need a flag...

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+	dnsRoot := fmt.Sprintf("client.endentity.%s.%s", input.Organization, randomLetters)
+	now := time.Now()
+	conf := &x509.Certificate{
+		SerialNumber:          big.NewInt(2019),
+		Subject:               subject,
+		NotBefore:             now.Add(-time.Minute), // grace of 1 min
+		NotAfter:              now.Add(input.Duration),
+		BasicConstraintsValid: true,
+		DNSNames:              []string{dnsRoot},
+		MaxPathLen:            0,
```
mastermanu, line 187: [Nit] MaxPathLen assignment unnecessary here since its default

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+							Name:  "organization-unit",
+							Usage: "The name of the organization unit (optional)",
+						},
+						&cli.StringFlag{
+							Name:     "duration",
+							Usage:    "The duration of the client certificate, for example: 1y30d (1 year and 30 days)",
+							Aliases:  []string{"d"},
+							Required: true,
+							Action: func(_ *cli.Context, v string) error {
+								if _, err := utils.ParseDuration(v); err != nil {
```
mastermanu, line 354: so this handles days?

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+									return fmt.Errorf("failed to parse duration: %v", err)
+								}
+								return nil
+							},
+						},
+						&cli.PathFlag{
+							Name:     CaCertificateFileFlagName,
```
mastermanu, line 361: the name of the command already has the word "ca" in it. Can we remove "ca" from here?

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+									return fmt.Errorf("duration cannot be more than: %s", maxCADuration)
+								}
+								if d <= minCADuration {
+									return fmt.Errorf("duration cannot be less than: %s", minCADuration)
+								}
+								return nil
+							},
+						},
+						&cli.PathFlag{
+							Name:     CaCertificateFileFlagName,
```
mastermanu, line 273: given these flags are for a command to generate ca certs, I think we can remove the word "ca/certificate authority" from them and just have it as --cert and --key

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+						&cli.BoolFlag{
+							Name:    "rsa-algorithm",
+							Aliases: []string{"rsa"},
+							Usage:   "Generate the certificate-authority using the RSA algorithm instead of ecdsa (optional)",
+						},
+					},
+					Action: func(ctx *cli.Context) error {
+						duration, err := utils.ParseDuration(ctx.String("duration"))
+						if err != nil {
+							return err
```
mastermanu, line 293: [Nit] wrap error message

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+								ctx.Path(CaPrivateKeyFileFlagName),
+							),
+						)
+						if err != nil || !yes {
+							return nil
+						}
+						err = ioutil.WriteFile(
+							ctx.Path(CaCertificateFileFlagName),
+							caPem,
+							0600,
```
mastermanu, line 315: [Nit] does CA certificate need same permissions as the private key? Being able to pass around the public key I think is fair game and this makes it a bit harder to do if we set 0600 for the public key as well as the private?

app/certificates.go
mastermanu: i guess because we are parsing string vs using duration flag

app/certificates.go
mastermanu: ok I see the annoyance when it comes to client cert generation and avoiding flag name collisions there

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+							Aliases:  []string{"org"},
+							Required: true,
+						},
+						&cli.StringFlag{
+							Name:  "organization-unit",
+							Usage: "The name of the organization unit (optional)",
+						},
+						&cli.StringFlag{
+							Name:     "duration",
+							Usage:    "The duration of the client certificate, for example: 1y30d (1 year and 30 days)",
```
mastermanu, line 350: are we checking that this doesn't CA expiration?

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+							Aliases:  []string{"d"},
+							Required: true,
```
mastermanu, line 352: Does it make sense to have default values for this? Or is it safer to just always ask user to supply it?

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+							Aliases:  []string{"ca-cert"},
+							Required: true,
+						},
+						&cli.PathFlag{
+							Name:     CaPrivateKeyFileFlagName,
+							Usage:    "The path of the file where the certificate-authority's private key is stored at",
+							Aliases:  []string{"ca-key"},
+							Required: true,
+						},
+						&cli.PathFlag{
```
mastermanu, line 372: do we need to make sure that a user cannot overwrite the files for ca certs =p

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+						if err != nil || !yes {
+							return nil
+						}
+						err = ioutil.WriteFile(
+							ctx.Path("certificate-file"),
+							certPem,
+							0600,
+						)
+						if err != nil {
+							return fmt.Errorf("failed to write ca certificate: %s", err)
```
mastermanu, line 424: error messages here and below refer to 'ca'

app/namespace.go
```diff
@@ -23,6 +24,7 @@ const (
 	namespaceRegionFlagName          = "region"
 	CaCertificateFlagName            = "ca-certificate"
 	CaCertificateFileFlagName        = "ca-certificate-file"
+	CaPrivateKeyFileFlagName         = "ca-private-key-file"
```
mastermanu, line 27: [Nit] worth aligning with tctl (except we use hyphens instead of '-')? Specifically do we need the word "private" here?

utils/duration.go
```diff
@@ -0,0 +1,214 @@
...
+	"d": uint64(24 * time.Hour),
+	"y": uint64(365 * 24 * time.Hour),
+}
+
+/*
+	A parser to parse time.Duration that can support days and year units.
+	Copied over from: https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/time/format.go;l=1589
+*/
+
+func ParseDuration(s string) (time.Duration, error) {
```
sergeybykov, line 117: Is there no OSS library to reuse for this instead of writing our own?

app/certificates.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+	caPrivateKey, err := x509.ParsePKCS8PrivateKey(pemBlock.Bytes)
+	if err != nil {
+		return nil, nil, false, fmt.Errorf("parsing ca private key failed: %s", err)
+	}
+	_, isRSA := caPrivateKey.(*rsa.PrivateKey)
+	return caCert, caPrivateKey, isRSA, nil
+}
+
+func generateCertificate(
```
sergeybykov, line 159: Nit: generateClientCertificate (for clarity)?

app/certificates.go
sergeybykov, line 350: I thought we wanted to limit duration to 1 year, no? the example of 1y30d clearly violates that and would be confusing.

sergeybykov (approved):
In general, LGTM minus the comments (primarily from @mastermanu )

app/certificates.go
anekkanti: fixed

app/certificates.go
anekkanti: thank you fixed.

app/certificates.go
anekkanti: I feel like its safer for the user to know explicitly that there is a expiration and to plan for it.

app/certificates.go
anekkanti: well there is only so much checking we can do.

app/certificates.go
anekkanti: thank you fixed.

cretz (approved):
After review over Zoom, this mostly LGTM with what was pointed out there. So long as we confirm each major SDK can connect w/ cert generated via this tool, I have nothing much to add.

app/certificates.go
```diff
@@ -169,61 +195,77 @@ func generateCertificate(
...
+	if input.Duration != 0 {
+		// a duration was provided by the user, validate it
+		notAfter = now.Add(input.Duration).UTC()
+		if notAfter.After(caCert.NotAfter.UTC()) {
+			return nil, nil, fmt.Errorf("duration of %s puts certificate's expiry after certificate authority's expiry %s by %s",
+				input.Duration, caCert.NotAfter.UTC().String(), notAfter.Sub(caCert.NotAfter.UTC()))
+		}
+	} else {
+		// set notAfter to ca's notAfter when duration is not set
+		notAfter = caCert.NotAfter.UTC()
```
mastermanu, line 221: i thought we were going to subtract like a day from this?

app/certificates.go
```diff
@@ -404,39 +430,73 @@ func NewCertificatesCommand() (CommandOut, error) {
...
-						return nil
 					},
 				},
 			},
 		},
 	}, nil
 }
+
+func checkPath(ctx *cli.Context, path string) (bool, error) {
+	if fi, err := os.Stat(path); !errors.Is(err, os.ErrNotExist) {
```
mastermanu, line 451: this is assuming that an error returned other than "not found" assumes the file exists? but that may also not be true?

app/certificates.go
```diff
@@ -404,39 +430,73 @@ func NewCertificatesCommand() (CommandOut, error) {
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to confirm: %w", err)
+	}
+	if !yes {
+		return nil
+	}
+	err = ioutil.WriteFile(certPath, cert, 0644)
+	if err != nil {
+		return fmt.Errorf("failed to write client certificate: %w", err)
```
mastermanu, line 492: is this same function used for both the CA and client certificate? If so, I'd remove the word "client" from these

mastermanu: maybe format with "typ" ?

mastermanu (approved):

app/certificates.go
mastermanu: tested with latest Java SDK client, PKCS8 seems to work

PR #197, opened 2023-06-01
Add export in tcld

alice-yin:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/namespace.go
```diff
@@ -57,6 +60,38 @@ var (
...
+	sinkNameFlag = &cli.StringFlag{
+		Name:     "sink-name",
+		Usage:    "sink name",
+		Required: true,
+	}
+	sinkEnabledFlag = &cli.BoolFlag{
+		Name:  "enabled",
+		Usage: "Whether the export sink is enabled",
+	}
+	sinkTypeFlag = &cli.StringFlag{
```
mastermanu, line 72: is this meant to be destination type? we might want to reword or indicate that. Also I assume we will only support limited set of values (in this case, only S3?)

app/namespace.go
```diff
@@ -57,6 +60,38 @@ var (
 		"us-east-1",
 		"us-west-2",
 	}
+	sinkNameFlag = &cli.StringFlag{
+		Name:     "sink-name",
+		Usage:    "sink name",
```
mastermanu, line 65: TODO: Probably need to enforce length and character limits on sink name on server side now that this is user assigned...

app/namespace.go
```diff
@@ -57,6 +60,38 @@ var (
...
+	sinkEnabledFlag = &cli.BoolFlag{
+		Name:  "enabled",
+		Usage: "Whether the export sink is enabled",
+	}
+	sinkTypeFlag = &cli.StringFlag{
+		Name:     "type",
+		Usage:    "The type of the export sink",
+		Required: true,
+	}
+	sinkAssumedRoleFlag = &cli.StringFlag{
```
mastermanu, line 77: is this the same terminology we are using for the role ARN in cloud ui/docs for audit log? We should probably try and align the same words

app/namespace.go
```diff
@@ -57,6 +60,38 @@ var (
...
+		Name:     "assumed-role",
+		Usage:    "The assumed role for the sink",
+		Required: true,
+	}
+	sinkDestinationArnFlag = &cli.StringFlag{
+		Name:     "destination-arn",
+		Usage:    "The destination ARN for the sink",
```
mastermanu, line 84: how is this going to work when we introduce GCP / other possible sinks here? I wonder if we need to have subcommands for different destination types so that users dont get over-whelmed by too many flags

app/namespace.go
```diff
@@ -978,6 +1013,168 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+							Name:    "create",
+							Aliases: []string{"c"},
+							Flags: []cli.Flag{
+								NamespaceFlag,
+								sinkNameFlag,
+								sinkEnabledFlag,
+								sinkTypeFlag,
+								sinkAssumedRoleFlag,
+								sinkDestinationArnFlag,
+								&cli.StringFlag{
```
mastermanu, line 1031: we should probably indicate the region must match that of the namespace here? or maybe not include the region and let the server derive it from the namespace?

app/namespace.go
```diff
@@ -978,6 +1013,168 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+							Name:    "update",
+							Aliases: []string{"u"},
+							Flags: []cli.Flag{
+								NamespaceFlag,
+								sinkNameFlag,
+								sinkEnabledFlag,
+								sinkTypeFlag,
+								sinkAssumedRoleFlag,
+								sinkDestinationArnFlag,
+								&cli.StringFlag{
```
mastermanu, line 1097: type and region are not "updatable" right?

alice-yin:
Add support in diff https://github.com/temporalio/tcld/pull/217

PR #199, opened 2023-06-09
Use url.Values when creating post request

tminusplus:
## What was changed

Use `url.Values` when creating HTTP POST requests in tcld.

## Why?

Ensures that the POST data is encoded to be URL safe.

## Checklist
- [x] Tested the login flow `tcld logout` and `tcld login`, and then verified that `tcld u list` returned a list of users.

### DIFF
--- a/app/login.go
+++ b/app/login.go
@@ -9,7 +9,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


mcbryde (approved):
Looks great, thanks for the quick turnaround.

PR #200, opened 2023-06-12
Add flag to only show export related commands

alice-yin:
## What was changed

In namespace.go, there is nothing change related to subCommands. subCommands is completely copy and paste. The change I made is from line 974-1005. So this is for only showing export commands when "export" feature flag enabled. 

1. tcld feature command

image

2. show export command only when enable-export is true

image

feature file will be json format, its content:

{"enable-export":true}

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/app/feature.go
+++ b/app/feature.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 
@@ -20,16 +21,16 @@ type FeatureFlag struct {
 	Value bool   `json:"Value"`
 }
 
-func getFeatureFlagJSON() ([]FeatureFlag, error) {
-	featureFlagConfigPath := getFeatureFlagConfigPath()
-	return getFeatureFlagConfig(featureFlagConfigPath)
+func getFeatureFlags() ([]FeatureFlag, error) {
+	featureFlagConfigPath := getFeatureFlagConfigFilePath()
+	return getFeatureFlagsFromConfigFile(featureFlagConfigPath)
 }
 
-func getFeatureFlagConfigPath() string {
+func getFeatureFlagConfigFilePath() string {
 	return filepath.Join(ConfigDirFlag.Value, featureflagFileName)
 }
 
-func getFeatureFlagConfig(featureFlagConfigPath string) ([]FeatureFlag, error) {
+func getFeatureFlagsFromConfigFile(featureFlagConfigPath string) ([]FeatureFlag, error) {
 	// create config file if not exist
 	if _, err := os.Stat(featureFlagConfigPath); err != nil {
 		if err := ioutil.WriteFile(featureFlagConfigPath, []byte("[]"), 0644); err != nil {
@@ -41,45 +42,45 @@ func getFeatureFlagConfig(featureFlagConfigPath string) ([]FeatureFlag, error) {
 		return nil, err
 	}
 
-	var jsonData []FeatureFlag
-	err = json.Unmarshal(content, &jsonData)
+	var featureFlags []FeatureFlag
+	err = json.Unmarshal(content, &featureFlags)
 	if err != nil {
 		return nil, err
 	}
 
-	return jsonData, nil
+	return featureFlags, nil
 }
 
-func toggle_feature(feature string) error {
-	featureFlagConfigPath := getFeatureFlagConfigPath()
-	return toggle_feature_save_to_path(feature, featureFlagConfigPath)
+func toggleFeature(feature string) error {
+	featureFlagConfigPath := getFeatureFlagConfigFilePath()
+	return toggleFeatureSaveToPath(feature, featureFlagConfigPath)
 }
 
-func toggle_feature_save_to_path(feature string, path string) error {
-	jsonData, err := getFeatureFlagConfig(path)
+func toggleFeatureSaveToPath(feature string, path string) error {
+	featureFlags, err := getFeatureFlagsFromConfigFile(path)
 
 	if err != nil {
 		return err
 	}
 
 	found := false
-	for i, featureflag := range jsonData {
+	for i, featureflag := range featureFlags {
 		if featureflag.Name == feature {
-			jsonData[i].Value = !featureflag.Value
+			featureFlags[i].Value = !featureflag.Value
 			found = true
-			println("Feature flag", feature, "is now", jsonData[i].Value)
+			println("Feature flag", feature, "is now", featureFlags[i].Value)
 		}
 	}
 
 	if !found {
-		jsonData = append(jsonData, FeatureFlag{
+		featureFlags = append(featureFlags, FeatureFlag{
 			Name:  feature,
 			Value: true,
 		})
 		println("Feature flag", feature, "is now", true)
 	}
 
-	output, err := json.Marshal(jsonData)
+	output, err := json.Marshal(featureFlags)
 
 	if err != nil {
 		return err
@@ -104,7 +105,7 @@ func NewFeatureCommand() (CommandOut, error) {
 					Aliases: []string{"te"},
 					Usage:   "switch export on/off",
 					Action: func(c *cli.Context) error {
-						return toggle_feature(ExportFeatureFlag)
+						return toggleFeature(ExportFeatureFlag)
 					},
 				},
 				{
@@ -112,18 +113,19 @@ func NewFeatureCommand() (CommandOut, error) {
 					Aliases: []string{"g"},
 					Usage:   "get all feature flags Value",
 					Action: func(c *cli.Context) error {
-						jsonData, err := getFeatureFlagJSON()
+						featureFlags, err := getFeatureFlags()
 
 						if err != nil {
 							return err
 						}
-						jsonString, err := json.Marshal(jsonData)
+
+						// MarshalIndent the feature flags into a pretty JSON
+						prettyJSON, err := json.MarshalIndent(featureFlags, "", "    ")
 						if err != nil {
-							return err
+							log.Fatalf("Failed to generate json: %s", err)
 						}
 
-						fmt.Println(string(jsonString))
-
+						fmt.Printf("%s\n", prettyJSON)
 						return nil
 					},
 				},
--- a/app/feature.go
+++ b/app/feature.go
@@ -20,6 +20,11 @@ type FeatureFlag struct {
 	Value bool   `json:"Value"`
 }
 
+func getFeatureFlagJSON() ([]FeatureFlag, error) {
+	featureFlagConfigPath := getFeatureFlagConfigPath()
+	return getFeatureFlagConfig(featureFlagConfigPath)
+}
+
 func getFeatureFlagConfigPath() string {
 	return filepath.Join(ConfigDirFlag.Value, featureflagFileName)
 }
@@ -107,8 +112,7 @@ func NewFeatureCommand() (CommandOut, error) {
 					Aliases: []string{"g"},
 					Usage:   "get all feature flags Value",
 					Action: func(c *cli.Context) error {
-						featureFlagConfigPath := getFeatureFlagConfigPath()
-						jsonData, err := getFeatureFlagConfig(featureFlagConfigPath)
+						jsonData, err := getFeatureFlagJSON()
 
 						if err != nil {
 							return err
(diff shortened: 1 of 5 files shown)


mastermanu (changes requested):
talked offline, we can use /.config/tcld and store feature flag data in there

app/feature.go
```diff
@@ -0,0 +1,80 @@
...
+		Usage:    "enable export commands",
+		Required: false,
+	}
+)
+
+func GetFeatureFlagConfigPath() string {
+	return ConfigDirFlag.Value + "/feature.json"
+}
+
+func GetFeatureFlagConfig(featureFlagConfigDir string) (map[string]bool, error) {
```
mastermanu, line 24: [Nit] featureFlagConfigDir --> featureFlagConfigPath

app/feature.go
```diff
@@ -0,0 +1,80 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/urfave/cli/v2"
+)
+
+var (
+	EnableExportFeatureFlag = &cli.BoolFlag{
+		Name:     "enable-export",
```
mastermanu, line 13: [Nit] Let's just call this `toggle-export` and it will just invert whatever value is already in the config map. Otherwise, the way this is, if we had multiple feature flags, a user has to remember to add the flags for the existing ones already enabled, which is confusing

app/feature.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return jsonData, nil
+}
+
+func NewFeatureCommand() (CommandOut, error) {
+	return CommandOut{
+		Command: &cli.Command{
```
mastermanu, line 48: is there a command that just prints out the currently enabled feature flags? Should we add it?

app/namespace.go
```diff
@@ -469,519 +875,134 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+						return c.renameSearchAttribute(ctx, n, existingName, newName)
 					},
 				},
 			},
 		},
-	}, nil
+	}
+
+	// ----------- Commands for private review feature, only available when feature flag turns on -----------
+	featureFlagConfigDir := GetFeatureFlagConfigPath()
```
mastermanu, line 975: [Nit] should we just have GetFeatureFlagConfig call GetFeatureFlagConfigPath internally?

app/namespace.go
```diff
@@ -469,519 +875,134 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+	jsonData, err := GetFeatureFlagConfig(featureFlagConfigDir)
+	if err != nil {
+		return CommandOut{}, err
+	}
+
+	// Export Command
+	if jsonData[EnableExportFeatureFlag.Name] {
```
mastermanu, line 982: [Nit] suggest separating tcld flag names and feature flag key values in the dictionary

app/namespace.go
```diff
@@ -469,519 +875,134 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+	command := &cli.Command{
+		Name:    "namespace",
+		Aliases: []string{"n"},
+		Usage:   "Namespace operations",
+		Before: func(ctx *cli.Context) error {
+			var err error
+			c, err = getNamespaceClientFn(ctx)
+			return err
+		},
+		Flags: []cli.Flag{
```
mastermanu, line 999: we dont need this EnableExportFeatureFlag here anymore, right?

mastermanu (changes requested):
left a few more comments.

app/feature.go
```diff
@@ -0,0 +1,80 @@
...
+			Name:    "feature",
+			Aliases: []string{"f"},
+			Usage:   "feature commands",
```
mastermanu, line 51: should we make this hidden?

alice-yin: Yes. I'm thinking this too.

app/namespace.go
alice-yin: It's due to feature test file. Want to give a temp file path in the test

app/feature.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/urfave/cli/v2"
+)
+
+var (
+	ExportFeatureFlag = "enable-export"
+)
+
+func GetFeatureFlagConfigPath() string {
+	return ConfigDirFlag.Value + "/feature.json"
```
mattkim, line 16: can you refactor to copy the login path logic:

```
configDir := ctx.Path(ConfigDirFlagName)
return filepath.Join(configDir, tokenFileName)
```

mattkim: Also does using the .Value notation work? I thought we usually use `ctx.Path` approach instead.

app/feature.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func GetFeatureFlagConfig(featureFlagConfigPath string) (map[string]bool, error) {
+	// create config file if not exist
+	if _, err := ioutil.ReadFile(featureFlagConfigPath); err != nil {
```
mattkim, line 21: we are using os.Stat to determine if files exist--I believe that is more standard?
```	tokenConfig := getTokenConfigPath(ctx)
	if _, err := os.Stat(tokenConfig); err != nil {
		// Skip if file does not exist
		if errors.Is(err, os.ErrNotExist) {
			return tokens, nil
		}
		return tokens, err
	}
```

mattkim: also should we just return early instead of creating the config? that way we can just create the config at creation time.

app/feature.go
```diff
@@ -0,0 +1,97 @@
...
+			return nil, err
+		}
+	}
+
+	content, err := ioutil.ReadFile(featureFlagConfigPath)
+	if err != nil {
+		return nil, err
+	}
+
+	var jsonData map[string]bool
```
mattkim, line 32: would prefer if this was a struct type so we can structure our data

app/feature.go
```diff
@@ -0,0 +1,97 @@
...
+						fmt.Println(ExportFeatureFlag, ":", jsonData[ExportFeatureFlag])
+
+						jsonString, err := json.Marshal(jsonData)
+						if err != nil {
+							return err
+						}
+
+						if err := ioutil.WriteFile(featureFlagConfigPath, jsonString, 0644); err != nil {
+							return err
+						}
```
mattkim, line 71: I feel like we could put this into a common helper function for future flags like api keys

app/feature.go
```diff
@@ -0,0 +1,97 @@
...
+						featureFlagConfigPath := GetFeatureFlagConfigPath()
+						jsonData, err := GetFeatureFlagConfig(featureFlagConfigPath)
+
+						if err != nil {
+							return err
+						}
+
+						for key, value := range jsonData {
+							fmt.Println(key, ":", value)
+						}
```
mattkim, line 89: i think if we have a dedicated struct for feature flags we can just print that out as json--we try to make every output json so it is machine parseable

app/namespace.go
```diff
@@ -469,519 +875,131 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
 		},
-	}, nil
+	}
+
+	// ----------- Commands for private review feature, only available when feature flag turns on -----------
+	featureFlagConfigDir := GetFeatureFlagConfigPath()
+	jsonData, err := GetFeatureFlagConfig(featureFlagConfigDir)
+	if err != nil {
+		return CommandOut{}, err
+	}
```
mattkim, line 979: can we move this into a single helper function so it's easier to use for api keys

app/feature.go
```diff
@@ -0,0 +1,97 @@
...
+func NewFeatureCommand() (CommandOut, error) {
+	return CommandOut{
+		Command: &cli.Command{
+			Name:    "feature",
+			Aliases: []string{"f"},
+			Usage:   "feature commands",
+			Hidden:  true,
+			Subcommands: []*cli.Command{
+				{
+					Name:    "toggle-export",
```
mattkim, line 50: actually looking at this again i wonder if it's better to offer an enable/disable option per feature flag?

in the future a feature flag might also contain a value--so i think structuring it as key/value makes sense to me.

app/feature.go
alice-yin: Yeah. It works.

app/feature.go
alice-yin: So the file will only contains feature flag and it's value:
{
"enable-export": true,
// in the future for other feature
"enable-XXX": false,
}

So this gives more flexibility for adding feature flag in the future. So I think Json would be fine?

app/feature.go
alice-yin: Yeah sure.

app/feature.go
alice-yin: toggle-export will be working in that way. I'm flipping the value all the time. Can find the detail on top.

app/feature.go
alice-yin: yes

app/namespace.go
alice-yin: Sure

app/feature.go
```diff
@@ -0,0 +1,133 @@
...
+	ExportFeatureFlag   = "enable-export"
+	featureflagFileName = "feature.json"
+)
+
+type FeatureFlag struct {
+	Name  string `json:"Name"`
+	Value bool   `json:"Value"`
+}
+
+func getFeatureFlagJSON() ([]FeatureFlag, error) {
```
mastermanu, line 23: [Nit] function is called getFeatureFlagJSON(), but this is just returning a slice of feature flags, so maybe we can just call it getFeatureFlags() ?

app/feature.go
```diff
@@ -0,0 +1,133 @@
...
+		if err := ioutil.WriteFile(featureFlagConfigPath, []byte("[]"), 0644); err != nil {
+			return nil, err
+		}
+	}
+	content, err := ioutil.ReadFile(featureFlagConfigPath)
+	if err != nil {
+		return nil, err
+	}
+
+	var jsonData []FeatureFlag
```
mastermanu, line 44: [Nit] rename variable to featureFlags. you can rename content to 'json data' if you'd like

app/feature.go
```diff
@@ -0,0 +1,133 @@
...
+	err = json.Unmarshal(content, &jsonData)
+	if err != nil {
+		return nil, err
+	}
+
+	return jsonData, nil
+}
+
+func toggle_feature(feature string) error {
```
mastermanu, line 53: [Nit] snake casing is not typically used in go? should we use camel case here?

app/feature.go
```diff
@@ -0,0 +1,133 @@
...
+	featureFlagConfigPath := getFeatureFlagConfigPath()
+	return toggle_feature_save_to_path(feature, featureFlagConfigPath)
+}
+
+func toggle_feature_save_to_path(feature string, path string) error {
+	jsonData, err := getFeatureFlagConfig(path)
```
mastermanu, line 59: [Nit] rename jsonData to featureFlags

app/feature.go
```diff
@@ -0,0 +1,133 @@
...
+					Name:    "get",
+					Aliases: []string{"g"},
+					Usage:   "get all feature flags Value",
+					Action: func(c *cli.Context) error {
+						jsonData, err := getFeatureFlagJSON()
+
+						if err != nil {
+							return err
+						}
+						jsonString, err := json.Marshal(jsonData)
```
mastermanu, line 120: [Nit] I think there is a PrettyPrintObj function you can use to pass in `jsonData` value here that does this for you already

app/feature.go
```diff
@@ -0,0 +1,133 @@
...
+					Action: func(c *cli.Context) error {
+						return toggle_feature(ExportFeatureFlag)
+					},
+				},
+				{
+					Name:    "get",
+					Aliases: []string{"g"},
+					Usage:   "get all feature flags Value",
+					Action: func(c *cli.Context) error {
+						jsonData, err := getFeatureFlagJSON()
```
mastermanu, line 115: [Nit] rename from jsonData to featureFlags

mastermanu (approved):

app/feature.go
alice-yin: Yess. I've got used to python..

alice-yin:
Somehow it asked me resolve all the conversations. So will do it manually.

PR #207, opened 2023-06-15
update namespace protogen

mattkim:
## What was changed

update with the latest namespace protogen

## Why?

to avoid any proto conflicts


yux0 (approved):
I am fine with this as long as team approve it.

mastermanu (approved):

PR #217, opened 2023-06-27
Add Export Support in Tcld

alice-yin:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/common.go
```diff
@@ -15,3 +24,18 @@ func IsFeatureEnabled(feature string) bool {
...
 	return false
 }
+
+func parseAssumedRole(assumedRole string) (string, string, error) {
+	var accountId, roleName string
+	re := assumedRolePattern
+	submatch := re.FindStringSubmatch(assumedRole)
+
+	if len(submatch) != 3 {
+		return "", "", fmt.Errorf("Invalid assumed role: %s", assumedRole)
```
mastermanu, line 34: [Nit] generally error strings in golang are not capitalized

app/common.go
```diff
@@ -15,3 +24,18 @@ func IsFeatureEnabled(feature string) bool {
 
 	return false
 }
+
+func parseAssumedRole(assumedRole string) (string, string, error) {
+	var accountId, roleName string
```
mastermanu, line 29: [Nit] golang convention for naming this variable would be accountID

app/namespace.go
```diff
@@ -56,6 +57,47 @@ var (
...
+		Usage: "Provide role arn for the IAM Role",
+	}
+	sinkAssumedRoleFlagRequired = &cli.StringFlag{
+		Name:     "role-arn",
+		Usage:    "Provide role arn for the IAM Role",
+		Required: true,
+	}
+	s3BucketFlagOptional = &cli.StringFlag{
+		Name:  "s3-bucket-name",
+		Usage: "Provide the name of an AWS S3 bucket that Temporal will send events to",
```
mastermanu, line 76: [Nit] we should replace the term "events" with "workflow histories" or something similar to that effect, right?

app/namespace.go
```diff
@@ -56,6 +57,47 @@ var (
...
 		"us-west-2",
 	}
+	sinkNameFlag = &cli.StringFlag{
+		Name:     "sink-name",
+		Usage:    "Provide a name for the export sink",
+		Required: true,
+	}
+	sinkAssumedRoleFlagOptional = &cli.StringFlag{
+		Name:  "role-arn",
+		Usage: "Provide role arn for the IAM Role",
```
mastermanu, line 67: [Nit] should we mention anything that suggests that the IAM Role has to be in the same account as the S3 bucket? Or is that too verbose for here?

app/namespace.go
```diff
@@ -82,6 +124,43 @@ func GetNamespaceClient(ctx *cli.Context) (*NamespaceClient, error) {
 	return NewNamespaceClient(ct, conn), nil
 }
 
+func (c *NamespaceClient) getExistingExportSink(ctx *cli.Context, nameSpaceName, sinkName string) (*sink.ExportSink, error) {
```
mastermanu, line 127: [Nit] nameSpaceName --> namespaceName

mastermanu: [Nit] getExistingExportSink --> getExportSink

app/namespace.go
```diff
@@ -82,6 +124,43 @@ func GetNamespaceClient(ctx *cli.Context) (*NamespaceClient, error) {
...
+
+	getResp, err := c.client.GetExportSink(c.ctx, getRequest)
+	if err != nil {
+		return nil, err
+	}
+	return getResp.Sink, nil
+}
+
+func (c *NamespaceClient) getExportSinkResourceVersionWithSink(ctx *cli.Context, sink *sink.ExportSink, nameSpaceName, sinkName string) (string, error) {
+	var resourceVersion string
```
mastermanu, line 141: [Nit], this is fine, but can also be written like so for brevity:

```
if ctx.String(ResourceVersionFlagName) != "" {
   return ctx.String(ResourceVersionFlagName)
	} 
   return sink.ResourceVersion, nil
```

app/namespace.go
```diff
@@ -82,6 +124,43 @@ func GetNamespaceClient(ctx *cli.Context) (*NamespaceClient, error) {
...
+	}
+
+	getResp, err := c.client.GetExportSink(c.ctx, getRequest)
+	if err != nil {
+		return nil, err
+	}
+	return getResp.Sink, nil
+}
+
+func (c *NamespaceClient) getExportSinkResourceVersionWithSink(ctx *cli.Context, sink *sink.ExportSink, nameSpaceName, sinkName string) (string, error) {
```
mastermanu, line 140: [Nit] this signature doesn't need to return an error

mastermanu, line 140: [Nit] getExportSinkResourceVersionWithSink --> selectExportSinkResourceVersion() ?

mastermanu: also why does this need namespaceName and sinkName as an argument?

app/namespace.go
```diff
@@ -82,6 +124,43 @@ func GetNamespaceClient(ctx *cli.Context) (*NamespaceClient, error) {
...
 
+func (c *NamespaceClient) getExistingExportSink(ctx *cli.Context, nameSpaceName, sinkName string) (*sink.ExportSink, error) {
+	getRequest := &namespaceservice.GetExportSinkRequest{
+		Namespace: nameSpaceName,
+		SinkName:  sinkName,
+	}
+
+	getResp, err := c.client.GetExportSink(c.ctx, getRequest)
+	if err != nil {
+		return nil, err
```
mastermanu, line 135: [Nit] might want to wrap this error with a friendly message

app/namespace.go
```diff
@@ -978,6 +1057,206 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+							return err
+						}
+
+						namespace := ctx.String(NamespaceFlagName)
+						ns, err := c.getNamespace(namespace)
+						if err != nil {
+							return fmt.Errorf("unable to get existing namespace: %v", err)
+						}
+
+						if ns.Namespace == "" {
```
mastermanu, line 1084: i dont think you need to do this check here btw. the api should return not found. should be easy to check though

app/namespace.go
```diff
@@ -978,6 +1057,206 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+					Action: func(ctx *cli.Context) error {
+						awsAccountID, roleName, err := parseAssumedRole(ctx.String(sinkAssumedRoleFlagRequired.Name))
+						if err != nil {
+							return err
+						}
+
+						namespace := ctx.String(NamespaceFlagName)
+						ns, err := c.getNamespace(namespace)
+						if err != nil {
+							return fmt.Errorf("unable to get existing namespace: %v", err)
```
mastermanu, line 1081: "unable to get namespace"

app/namespace.go
```diff
@@ -978,6 +1057,206 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+						request := &namespaceservice.CreateExportSinkRequest{
+							Namespace: namespace,
+							Spec: &sink.ExportSinkSpec{
+								Name:            ctx.String(sinkNameFlag.Name),
+								Enabled:         true,
+								DestinationType: sink.EXPORT_DESTINATION_TYPE_S3,
+								S3Sink: &sink.S3Spec{
+									RoleName:     roleName,
+									BucketName:   ctx.String(s3BucketFlagRequired.Name),
+									Region:       ns.Spec.Region,
```
mastermanu, line 1097: interesting that we take this as an argument right now (might be my mistake when I added this to the proto). One other option is to (at least currently) not take this as an argument and only take it later if we support cross-region export

app/namespace.go
```diff
@@ -978,6 +1057,206 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+				{
+					Name:    "get",
+					Aliases: []string{"g"},
+					Usage:   "Get export sink",
+					Flags: []cli.Flag{
+						NamespaceFlag,
+						sinkNameFlag,
+					},
+					Action: func(ctx *cli.Context) error {
+						request := &namespaceservice.GetExportSinkRequest{
```
mastermanu, line 1122: this seems like you can just call getExistingExportSink here instead and avoid code duplication?

app/namespace.go
```diff
@@ -978,6 +1057,206 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+						}
+
+						deleteRequest := &namespaceservice.DeleteExportSinkRequest{
+							Namespace:       nameSpaceName,
+							SinkName:        sinkName,
+							ResourceVersion: resourceVersion,
+							RequestId:       ctx.String(RequestIDFlagName),
+						}
+
+						detelteResp, err := c.client.DeleteExportSink(c.ctx, deleteRequest)
```
mastermanu, line 1160: [Nit] typo `detelteResp`

app/namespace.go
```diff
@@ -978,6 +1057,206 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+						RequestIDFlag,
+					},
+					Action: func(ctx *cli.Context) error {
+						awsAccountID, roleName, err := parseAssumedRole(ctx.String(sinkAssumedRoleFlagRequired.Name))
+						if err != nil {
+							return err
+						}
+
+						namespace := ctx.String(NamespaceFlagName)
+						ns, err := c.getNamespace(namespace)
```
mastermanu, line 1079: why do we check getNamespace here, but not for update,delete and get? I think the API should handle if the namespace doesn't exist, right?

app/namespace.go
```diff
@@ -978,6 +1057,206 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+						if err != nil {
+							return err
+						}
+
+						resourceVersion, err := c.getExportSinkResourceVersionWithSink(ctx, sink, nameSpaceName, sinkName)
+						if err != nil {
+							return err
+						}
+
+						if !ctx.IsSet(sinkEnabledFlag.Name) && !ctx.IsSet(sinkAssumedRoleFlagOptional.Name) && !ctx.IsSet(s3BucketFlagOptional.Name) && !ctx.IsSet(kmsArnFlag.Name) {
```
mastermanu, line 1218: what does it mean to update the kmsArnFlag? We should probably update the description to indicate that this is ONLY used for template generation?

Also what does the experience look like if customer tries to write to a s3 bucket where the role doesn't have appropriate KMS permissions? is that status/error message returned useful? we might want to test this scenario

app/namespace.go
```diff
@@ -978,6 +1057,206 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+							fmt.Println("nothing to update")
+							return nil
+						}
+
+						if ctx.IsSet(sinkEnabledFlag.Name) {
```
mastermanu, line 1223: if sink.spec.Enabled value is already equal to sinkEnabledFlag.Value, then technically that would also not be a change, right? Same if all of their flags supply already existing values?

app/namespace_test.go
```diff
@@ -1296,3 +1306,354 @@ func (s *NamespaceTestSuite) TestDelete() {
 	}, nil).Times(1)
 	s.NoError(s.RunCmd("namespace", "delete", "--namespace", "ns1"))
 }
+
+func (s *NamespaceTestSuite) TestCreateExportSink() {
+	ns := "namespace"
+	type morphGetResp func(*namespaceservice.GetNamespaceResponse)
+	type morphCreateSinkReq func(*namespaceservice.CreateExportSinkRequest)
+
+	tests := []struct {
```
mastermanu, line 1315: [Nit] add test case name as argument to make it clear what the intent of each of these table tests is

mastermanu: or if we are expecting an error, consider checking if expected error string is returned

app/namespace_test.go
```diff
@@ -1296,3 +1306,354 @@ func (s *NamespaceTestSuite) TestDelete() {
...
+			if tc.expectErr {
+				s.Error(err)
+			} else {
+				s.NoError(err)
+			}
+		})
+	}
+}
+
+func (s *NamespaceTestSuite) TestListExportSink() {
```
mastermanu, line 1491: [nit] TestListExportSinks

mastermanu (approved):

app/namespace.go
alice-yin: Yeah. I think in this step is what user needs to create the role beforehand. So we'll need to provide a documentation on how to set the role up and noted the role need to be on the same account of S3

app/namespace.go
alice-yin: This is copy from UI, will ask UI to change as well. Will use "closed workflow histories"

app/namespace.go
alice-yin: remove error and sinkName. I've done some refactoring before and forgot to remove those unneeded input parameter.

PR #225, opened 2023-07-27
Add ability to set common-name on leaf certificates generated via tcld

mastermanu:


app/certificates_test.go
```diff
@@ -50,6 +50,8 @@ func (s *CertificatesTestSuite) TestCertificateGenerateCore() {
 		caValidityPeriod        time.Duration
 		endEntityValidityPeriod time.Duration
 		organization            string
+		organizationalUnit      string
```
mastermanu, line 53: rename to organizationUnit

jumbogo (approved):

PR #230, opened 2023-08-17
add missing kms arn flag in create export sink

alice-yin:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/app/namespace.go
+++ b/app/namespace.go
@@ -87,7 +87,7 @@ var (
 	}
 	kmsArnFlag = &cli.StringFlag{
 		Name:  "kms-arn",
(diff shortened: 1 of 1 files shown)


app/namespace.go
```diff
@@ -1161,6 +1161,7 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 						sinkAssumedRoleFlagRequired,
 						s3BucketFlagRequired,
 						RequestIDFlag,
+						kmsArnFlag,
```
alice-yin, line 1164: Already use it on line 1188. Just forgot to provide it as the input.

shivam-ajmera (approved):

PR #258, opened 2023-09-26
Add refresh token support for logins

tminusplus:
## What was changed

This adds refresh tokens to tcld by switching to the `golang.org/x/oauth2` which supports device codes and refresh tokens.

Other changes in this PR:
- Noticed that other tests were reading the actual tcld config folder. And would panic if it does not exist. So added `NewTestApp` which sets the config dir to a temporary directory + other convenience settings.
- Removed LoginService to reduce coupling in code, so that `connection.go` can be tested without invoking a login.
- Removed mocks around LoginService, to ensure the tcld code is exercised in tests.
- Added making the config directory, if it doesn't exist, in `feature.go` to avoid nil panics in testing.
- Removed the `Before` code in the `NewLoginCommand`, as nothing was done with the loaded login config.
- Removed the use of `os.RemoveAll`, to ensure a bug does not remove directories on a user's computer when the intent was to remove a single file.

## Why?

Allows us to shorten token access expiration times.

## Checklist
- [x] Tested logging in and making a request.
- [x] Tested logging in on an older tcld, updating, and then making a request.
- [x] Tested logging in on an older tcld, updating, re-logging in, and verified the config is in the new format.
- [x] Tested waiting for the access token to expire, and then making a request to get a new refresh token.
- [x] Added unit tests around refresh token support and the use of the `golang.org/x/oauth2` library.
- [x] Tested login flows with refresh tokens disabled server-side.


app/credentials/oauth/oauth.go
```diff
@@ -42,27 +38,30 @@ func NewCredential(accessToken string, opts ...Option) (Credential, error) {
...
 	}
 
+	token, err := c.source.Token()
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve token from token source: %w", err)
+	}
+
 	return map[string]string{
-		Header: c.token(),
+		credentials.AuthorizationHeader: fmt.Sprintf("%s %s", token.Type(), token.AccessToken),
```
mattkim, line 59: are we supposed to return the AccessToken or RefreshToken here?

app/login.go
```diff
@@ -171,35 +148,16 @@ func (c *LoginClient) login(ctx *cli.Context, domain string, audience string, cl
...
 
-	tokenRespJson, err := FormatJson(tokenResp)
+	tokenRespJson, err := FormatJson(token)
 	if err != nil {
 		return err
 	}
-	fmt.Println("Successfully logged in!")
 
 	// Save token info locally
 	return c.loginService.WriteToConfigFile(getTokenConfigPath(ctx), tokenRespJson)
```
mattkim, line 163: Basic refresh token question from me: Is there a way to never write the access token locally and only write the refresh token? In my brain part of the value of refresh token is that we never need to store the access token so it can't be leaked? Or is it that we fetch the the access token once up front and then subsequently send refresh tokens until they expire and need to be fetched again? To me that is less valuable since we will need to send the access token over the wire every 15s anyways (if that's the refresh token expiry)

app/credentials/oauth/oauth.go
tminusplus: The access token, since that is the bearer token the API is looking for to authenticate.

app/login.go
tminusplus: Ah its the opposite - the refresh token is now more valuable than the access token since it has a longer lifetime and can be used in an exchange for a new access token. Ideally we store this information on the OS' keyring, like using this [library](https://github.com/99designs/keyring), to prevent leaking the token locally.

We will need to store the access token somewhere for use with requests, and the expiration should be in the minutes territory (lets say 15m for illustration).

When the access token is past its expiry, the oauth2 library will exchange the refresh token for a new refresh token and access token before making the gRPC request. The refresh token is exchanged as added protection because it prevents refresh token re-use if credential theft ever occurs. Re-using an invalidated refresh token will also invalidate all refresh tokens handed out to the user, its called [refresh token rotation](https://auth0.com/docs/secure/tokens/refresh-tokens/refresh-token-rotation).

I still need to validate all this behavior in a test environment before merging, but with out current setup I did verify this works.

app/connection_test.go
```diff
@@ -40,6 +42,8 @@ func (s *testServer) GetRequestStatus(ctx context.Context, req *requestservice.G
 	md, _ := metadata.FromIncomingContext(ctx)
 	s.receivedMD = md.Copy()
 
+	fmt.Printf("Received md is %#v\n", md)
```
shakeelrao, line 45: debug log :^)

app/credentials/oauth/oauth.go
```diff
@@ -42,27 +38,30 @@ func NewCredential(accessToken string, opts ...Option) (Credential, error) {
...
 }
 
 func (c Credential) RequireTransportSecurity() bool {
 	return !c.allowInsecureTransport
 }
 
-func (c Credential) token() string {
-	return "Bearer " + c.accessToken
-}
+var _ grpccreds.PerRPCCredentials = (*Credential)(nil)
```
shakeelrao, line 67: noob q: what does this do?

app/connection_test.go
```diff
@@ -191,14 +192,11 @@ func (s *ServerConnectionTestSuite) TestGetServerConnection() {
...
-				authHeader := getHeaderValue(md, apikey.AuthorizationHeader)
-				s.Equal("Bearer "+testAPIKey, authHeader)
-			} else {
-				token := getHeaderValue(md, oauth.Header)
-				s.Equal("Bearer "+testAccessToken, token)
-			}
+			auth := strings.SplitN(getHeaderValue(md, credentials.AuthorizationHeader), " ", 2)
+			require.Len(s.T(), auth, 2)
+
+			s.Equal(strings.ToLower(credentials.AuthorizationBearer), strings.ToLower(auth[0]))
```
shakeelrao, line 198: fftI: can use `strings.EqualFold` here.

app/credentials/header.go
```diff
@@ -0,0 +1,6 @@
+package credentials
+
+const (
+	AuthorizationHeader = "Authorization"
+	AuthorizationBearer = "Bearer"
```
shakeelrao, line 5: `BearerAuthorization` seems clearer to me, wdyt?  

(feel free to ignore)

tminusplus: Hmm I'm not sure it is much clearer, but also agree that `AuthorizationBearer` is not great. What about `BearerTokenType` or `TokenTypeBearer`?

app/credentials/oauth/oauth.go
tminusplus: Its a good question. It ensures that the structure `Credential` that we wrote satisfies the `grpccreds.PerRPCCredentials` interface at compile time. Its called an [interface check](https://go.dev/doc/effective_go#blank_implements) in Go, and is a consequence of Go interfaces being implicit.

They made them implicit (meaning no `implements` keyword) to decouple the definition of an interface from the implementation. It also allows a structure to implement many interfaces, and encourages a design idiom in Go where interfaces are small and well-defined. And other things.

Great blog post on [why interfaces are implicit](https://dave.cheney.net/2016/08/20/solid-go-design) from Dave Cheney.

app/credentials/header.go
shakeelrao: `BearerTokenType` sgtm!

go.mod
```diff
@@ -29,11 +30,12 @@ require (
 	go.uber.org/atomic v1.7.0 // indirect
 	go.uber.org/dig v1.16.1 // indirect
 	go.uber.org/zap v1.23.0 // indirect
-	golang.org/x/crypto v0.7.0 // indirect
-	golang.org/x/net v0.8.0 // indirect
-	golang.org/x/sys v0.6.0 // indirect
-	golang.org/x/text v0.8.0 // indirect
+	golang.org/x/crypto v0.14.0 // indirect
+	golang.org/x/net v0.16.0 // indirect
```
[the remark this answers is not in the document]
tminusplus: Will handle in https://github.com/temporalio/tcld/pull/263

shakeelrao (approved):
🚀

app/connection.go
```diff
@@ -77,29 +77,24 @@ func defaultDialOptions(c *cli.Context, addr *url.URL) ([]grpc.DialOption, error
...
 	}
 
-	if len(tokens.AccessToken) > 0 {
-		return oauth.NewCredential(
-			tokens.AccessToken,
-			oauth.WithInsecureTransport(insecure),
-		)
-	}
-
-	// Use no credentials for this connection.
```
mastermanu, line 103: was there any scenario where we wanted to use "no credentials" here? (e.g. development or w/e)

app/logout.go
```diff
@@ -15,17 +18,21 @@ func NewLogoutCommand(c *LoginClient) (CommandOut, error) {
 			disablePopUpFlag,
 		},
 		Action: func(ctx *cli.Context) error {
-			if err := c.loginService.DeleteConfigFile(getTokenConfigPath(ctx)); err != nil {
+			configDir := ctx.Path(ConfigDirFlagName)
```
mastermanu, line 21: should we still try to logout even in this error case?

app/token_config.go
```diff
@@ -0,0 +1,143 @@
...
+
+	data, err := os.ReadFile(tokenConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read login config: %w", err)
+	}
+
+	var config *TokenConfig
+	if err := json.Unmarshal(data, &config); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal login config: %w", err)
+	} else if config.OAuthToken == nil {
```
mastermanu, line 53: is expiry checked elsewhere?

mastermanu (approved):

app/connection.go
tminusplus: There is no scenario, I just put this there a while back to pass nothing if nothing was specified

app/logout.go
tminusplus: Oh yes we should.

PR #260, opened 2023-10-10
add validate export sink in tcld

alice-yin:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


jumbogo (approved):
This looks good to me.  Just curious, is there a way to enforce the customer to apply the policy document or cloudformation stack run when using tcld?

alice-yin:
> This looks good to me. Just curious, is there a way to enforce the customer to apply the policy document or cloudformation stack run when using tcld?

I haven't looked at in tcld about the cloudformation stack. In the user's doc, we ask user to set up role manually beforehand.

PR #261, opened 2023-10-10
add a flag to specify codec server for tcld create namespace

jumbogo:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/codecserver.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	return &namespace.CodecServerPropertySpec{
+		Endpoint:           codec.Endpoint,
+		PassAccessToken:    codec.PassAccessToken,
+		IncludeCredentials: codec.IncludeCredentials,
+	}, nil
+}
+
+func validateCodec(codec codecServer) error {
+	if codec.Endpoint == "" {
```
mastermanu, line 36: [Nit] should we do any other basic validation on the endpoint semantics itself? or not worth it?

app/namespace.go
```diff
@@ -429,6 +430,11 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 					Usage:   fmt.Sprintf("Flag can be used multiple times; value must be \"email=permission\"; valid permissions are: %v", getNamespacePermissionTypes()),
 					Aliases: []string{"p"},
 				},
+				&cli.StringFlag{
+					Name:    codecServerSpecJsonFlagName,
+					Usage:   `JSON that defines the codec server property passed to the namespace. Sample JSON: { "codecserver": { "endPoint": "https://test-endpoint.com", "PassAccessToken": true, "IncludeCredentials": false } }`,
```
mastermanu, line 435: can we run this by jonathan/abhinav? I think from a user experience its easier to just take three separate flags than to require the user to construct the JSON themselves

app/codecserver.go
jumbogo: the most important is https, do you have suggestions on other basic validations?

app/namespace.go
jumbogo: Changed to use separate flags.

app/namespace.go
```diff
@@ -89,6 +91,21 @@ var (
...
 		Usage: "The page token for list operations",
 	}
+	codecIncludeCredentialsFlag = &cli.BoolFlag{
+		Name:    codecIncludeCredentialsFlagName,
+		Usage:   "Include cross-origin credentials",
+		Aliases: []string{"ic"},
+	}
+	codecPassAccessTokenFlag = &cli.BoolFlag{
+		Name:    codecPassAccessTokenFlagName,
+		Usage:   "Pass the user access token with the remote endpoint",
```
mastermanu, line 101: [Nit] with --> to, right?

app/namespace.go
```diff
@@ -500,6 +520,20 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+				codecEndpoint := ctx.String(codecEndpointFlagName)
+				// codec server spec is optional, if specified, we need to create the spec and pass along to the API
+				if codecEndpoint != "" {
+					err = validateCodecInput(codecEndpoint)
+					if err != nil {
+						return err
+					}
+					n.Spec.CodecSpec = &namespace.CodecServerPropertySpec{
+						Endpoint:           codecEndpoint,
+						PassAccessToken:    ctx.Bool(codecPassAccessTokenFlagName),
```
mastermanu, line 532: if someone sets one of these two flags w/o setting the endpoint, should we fail?

app/namespace.go
```diff
@@ -1395,6 +1426,13 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 	return CommandOut{Command: command}, nil
 }
 
+func validateCodecInput(codecEndpoint string) error {
```
mastermanu, line 1429: [Nit] validateCodecEndpoint

app/codecserver.go
mastermanu: https://golang.cafe/blog/how-to-validate-url-in-go.html

mastermanu (approved):

app/namespace.go
jumbogo: Renamed

app/namespace.go
jumbogo: added error out in this case

app/namespace.go
jumbogo: Done

PR #268, opened 2023-10-20
Add tcld for create global namespace

yux0:
## What was changed
Add tcld for create global namespace

## Why?
Add tcld for create global namespace
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/app/namespace.go
+++ b/app/namespace.go
@@ -413,7 +413,7 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 				},
 				&cli.StringSliceFlag{
 					Name:     namespaceRegionFlagName,
-					Usage:    "Create namespace in specified regions; see 'tcld account list-regions' to get a list of available regions for your account",
+					Usage:    "Create namespace in specified regions; if multiple regions are selected, the first one will be the active region. See 'tcld account list-regions' to get a list of available regions for your account",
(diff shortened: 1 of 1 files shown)


app/namespace.go
```diff
@@ -459,6 +461,17 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 				n.Spec = &namespace.NamespaceSpec{
 					Region: ctx.String(namespaceRegionFlagName),
 				}
+				regions := ctx.StringSlice(namespaceRegionFlagName)
+				if len(regions) == 0 {
+					return fmt.Errorf("namespace region is required")
+				}
+				if len(regions) > 2 {
+					return fmt.Errorf("namespace can only be sepecified up to 2 regions")
```
wxing1292, line 469: sepecified -> replicated

app/namespace.go
```diff
@@ -459,6 +461,17 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+				}
+				n.Spec = &namespace.NamespaceSpec{
+					Region:         regions[0],
```
wxing1292, line 472: 1st region is the active region? we probably should make it explicit?

wxing1292 (approved):

PR #271, opened 2023-10-26
[CLD-1637] support API key duration in days

shakeelrao:
**Problem**
The `duration` flag for API key creation uses [time.ParseDuration](https://pkg.go.dev/time#ParseDuration), which only supports strings of the form `2h45m` (where hour is the highest level of granularity). However, the UI for API key creation only supports days, which is a mismatch in UX.

**Solution**
Switch to `utils.ParseDuration` which extends `time.ParseDuration` to accept years and days.

**Result**
Clients can specify API key expiration in days.

### DIFF
--- a/app/apikey.go
+++ b/app/apikey.go
@@ -210,7 +210,7 @@ func NewAPIKeyCommand(getAPIKeyClientFn GetAPIKeyClientFn) (CommandOut, error) {
 							}
 							d, err := parseDuration(expiryPeriod)
 							if err != nil {
-								return fmt.Errorf("failed to parse duration: %v", err)
+								return fmt.Errorf("failed to parse duration: %w", err)
 							}
 							if d == 0 {
 								return fmt.Errorf("no expiry was set")
--- a/app/apikey.go
+++ b/app/apikey.go
@@ -212,6 +212,9 @@ func NewAPIKeyCommand(getAPIKeyClientFn GetAPIKeyClientFn) (CommandOut, error) {
 							if err != nil {
 								return fmt.Errorf("failed to parse duration: %v", err)
 							}
+							if d == 0 {
+								return fmt.Errorf("no expiry was set")
+							}
 							e := time.Now().UTC().Add(d)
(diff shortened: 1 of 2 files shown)


app/apikey.go
```diff
@@ -122,6 +125,40 @@ func (s *APIKeyClient) deleteAPIKey(
...
+func parseDuration(s string) (time.Duration, error) {
+	var d time.Duration
+	durationString := s
+	parts := strings.Split(s, "d")
+	if len(parts) == 2 {
+		days, err := strconv.ParseInt(parts[0], 10, 32)
+		if err != nil {
+			return d, fmt.Errorf("time: invalid duration \"%s\"", s)
+		}
+		if days < 0 {
```
tminusplus, line 138: What happens if the user specifies `0d`?

app/apikey.go
```diff
@@ -167,11 +204,18 @@ func NewAPIKeyCommand(getAPIKeyClientFn GetAPIKeyClientFn) (CommandOut, error) {
...
 						if expiry == nil || expiry.IsZero() {
-							expiryPeriod := ctx.Duration("duration")
-							if expiryPeriod == 0 {
+							expiryPeriod := ctx.String("duration")
+							if expiryPeriod == "" {
+								return fmt.Errorf("no expiry was set")
+							}
+							d, err := parseDuration(expiryPeriod)
+							if err != nil {
+								return fmt.Errorf("failed to parse duration: %v", err)
```
tminusplus, line 213: Can be good to use `%w` so the error is [wrapped](https://bitfieldconsulting.com/golang/wrapping-errors).

tminusplus (approved):
Couple small comments otherwise :rocket:

app/apikey.go
shakeelrao: we'd return 0 i believe. technically `0d24h` works too...

shakeelrao: i added an addition check in the `Action` function to reject durations of 0. wdyt?

PR #275, opened 2023-10-31
feat: add env var for tcld config dir

sagikazarmark:
## What was changed
Added an environment variable for changing the configuration directory

## Why?
I may need to access different Temporal orgs and setting an env var is easier than passing a config flag every time.


tminusplus (approved):
Thank you for the contribution 🚀

PR #280, opened 2023-11-13
Move Export command out of private preview command

alice-yin:
## What was changed

Remove Export out of private preview. 

Plan to have the change in and cut the tcld release before public preview. 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/app/feature.go
+++ b/app/feature.go
@@ -11,10 +11,11 @@ import (
 )
 
 var (
-	ExportFeatureFlag   = "enable-export"
 	APIKeyFeatureFlag   = "enable-apikey"
 	featureflagFileName = "feature.json"
 )
+
+var supportFeatureFlags = []string{APIKeyFeatureFlag}
 
 type FeatureFlag struct {
 	Name  string `json:"Name"`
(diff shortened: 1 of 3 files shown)


app/namespace.go
```diff
@@ -1176,257 +1176,254 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 		},
 	}
 
-	// ----------- Commands for private review feature, only available when feature flag turns on -----------
-	// Export Command
-	if IsFeatureEnabled(ExportFeatureFlag) {
```
alice-yin, line 1181: The only change is remove 

`if IsFeatureEnabled(ExportFeatureFlag) {`

Rest are formatting.

app/feature.go
```diff
@@ -100,14 +100,6 @@ func NewFeatureCommand() (CommandOut, error) {
 			Usage:   "feature commands",
 			Hidden:  true,
 			Subcommands: []*cli.Command{
-				{
-					Name:    "toggle-export",
-					Aliases: []string{"te"},
-					Usage:   "switch export on/off",
```
mastermanu, line 106: can you remove export reference from here too? https://github.com/temporalio/tcld/blob/6196555f9a9d1c5aab65880858efe9700a554124/app/feature.go#L14

app/feature.go
```diff
@@ -100,14 +100,6 @@ func NewFeatureCommand() (CommandOut, error) {
...
-					Action: func(c *cli.Context) error {
-						return toggleFeature(ExportFeatureFlag)
```
mastermanu, line 108: `func getFeatureFlagsFromConfigFile(featureFlagConfigPath string) ([]FeatureFlag, error) { `

For the above function, if an older client has the export feature flag in their config, will the "get" command still print out export as a feature flag? If so, should we add code to protect against printing garbage values?

mastermanu (approved):

app/feature.go
alice-yin: Will skip printing it if we don't have this feature flag any more.

PR #281, opened 2023-11-15
tcld: Add gcp validate sink API for Export

alice-yin:
## What was changed

As title. Set up the first Validate sink api so we could start testing around GCP sink. 

image

image

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/namespace.go
```diff
@@ -108,6 +108,16 @@ var (
 		Usage:   "The codec server endpoint to decode payloads for all users interacting with this Namespace, must be https",
 		Aliases: []string{"e"},
 	}
+	serviceAccountPrincipalFlag = &cli.StringFlag{
+		Name:     "service-account-principal",
+		Usage:    "service account that could access to the sink",
```
mastermanu, line 113: [Nit] service account that has access to the sink ?

app/namespace.go
```diff
@@ -1425,6 +1435,66 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 		},
 	})
 
+	// Note: the following commands are depend on previous commands. Please don't add more commands in the middle
```
mastermanu, line 1438: this seems somewhat confusing / brittle. Is there a better way to programmatically search and find the export command we care about by maybe checking the .Name field?

app/namespace.go
```diff
@@ -1425,6 +1435,66 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+	exportCommand := subCommands[len(subCommands)-1]
+
+	// --- GCP Sink Related Commands ---
+	// Will move under export command once GCP sink is under public preview
+	if IsFeatureEnabled(GCPSinkFeatureFlag) {
+		exportCommand.Subcommands = append(exportCommand.Subcommands, &cli.Command{
+			Name:    "gcp-sink",
```
mastermanu, line 1445: so how does a user use this? `tcld es gcp validate?`

mastermanu: also, whats the future facing approach for this?

we currently  have `tcld n es validate` 
is the right thing to move the current validate under a new validate function and have s3? e.g.

`tcld n es validate s3 <flags>` and then `tcld n es validate gcs <flags>`

mastermanu: i think we should just make that above change now

app/namespace.go
```diff
@@ -1425,6 +1435,66 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+							},
+						}
+
+						_, err = c.client.ValidateExportSink(c.ctx, validateRequest)
+
+						if err != nil {
+							return fmt.Errorf("validation failed with error %v", err)
+						}
+
+						fmt.Println("Validate test file can be written to the sink successfully")
```
mastermanu, line 1490: [Nit] re-phrase/re-write: "Temporal Cloud was able to write test data to the sink"

mastermanu (approved):

app/namespace.go
alice-yin: Cool

app/namespace.go
alice-yin: So I actually think another way of handling this. use `tcld n es s3 validate`

PR #289, opened 2023-11-29
Add CRUD Export gcp sink in tcld

alice-yin:
## What was changed

The new command will be like following. 

image

image

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/namespace.go
```diff
@@ -108,16 +108,28 @@ var (
...
 		Name:     "gcs-bucket",
-		Usage:    "gcs bucket of the sink",
+		Usage:    "GCS bucket of the sink",
 		Required: true,
 	}
+	enableCmekFlag = &cli.BoolFlag{
+		Name:  "enable-cmek",
+		Usage: "Enable customized encryption.  Note: If enable cmek needs to be updated, user should create the service account with cmek or modify the created service account accordingly. Provided it as part of the input won't help",
+	}
+	gcsBucketFlagOptional = &cli.StringFlag{
```
mastermanu, line 129: [Nit] put optional and required flag definitions together for gcs bucket?

app/namespace.go
```diff
@@ -108,16 +108,28 @@ var (
...
+	}
+	gcsBucketFlagRequired = &cli.StringFlag{
 		Name:     "gcs-bucket",
-		Usage:    "gcs bucket of the sink",
+		Usage:    "GCS bucket of the sink",
 		Required: true,
 	}
+	enableCmekFlag = &cli.BoolFlag{
+		Name:  "enable-cmek",
+		Usage: "Enable customized encryption.  Note: If enable cmek needs to be updated, user should create the service account with cmek or modify the created service account accordingly. Provided it as part of the input won't help",
```
mastermanu, line 127: [Nit] the last sentence of this usage statement doesn't really provide any extra information/guidance. can we remove this?

let's also replace user "should" create the service account with cmek with user "must"

app/namespace.go
```diff
@@ -1193,6 +1225,84 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 		Aliases: []string{"es"},
 	}
 
+	export_general_commands := []*cli.Command{
```
mastermanu, line 1228: [Nit] use camel case?

app/namespace.go
```diff
@@ -1193,6 +1225,84 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+					return err
+				}
+
+				return PrintProto(deleteResp.RequestStatus)
+			},
+		},
+		{
+			Name:    "list",
+			Aliases: []string{"l"},
+			Usage:   "List export sink",
```
mastermanu, line 1283: [Nit] List export sinks

app/namespace.go
```diff
@@ -1312,60 +1405,125 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+					if err != nil {
+						return err
+					}
+
+					if !isEnabledChange && !c.isAssumedRoleChange(ctx, sink) && !c.isKmsArnChange(ctx, sink) && !c.isS3BucketChange(ctx, sink) {
+						fmt.Println("nothing to update")
+						return nil
+					}
+
+					if isEnabledChange {
```
mastermanu, line 1440: the behavior here suggests the flag is more of a "toggle" than an "enable?" Can we rename the flag and its usage to make this more clear?

app/namespace.go
```diff
@@ -1312,60 +1405,125 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
 					}
 
-					return PrintProto(deleteResp.RequestStatus)
+					return PrintProto(resp.RequestStatus)
 				},
 			},
+		},
+	}
+
+	export_gcs_commands := &cli.Command{
```
mastermanu, line 1479: [Nit] use camel case here for variable name

app/namespace.go
```diff
@@ -1396,7 +1554,7 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 						return err
 					}
 
-					if !isEnabledChange && !c.isAssumedRoleChange(ctx, sink) && !c.isKmsArnChange(ctx, sink) && !c.isS3BucketChange(ctx, sink) {
+					if !isEnabledChange && !c.isSAPrincipalChange(ctx, sink) && !c.isEnableCmekChange(ctx, sink) && !c.isGCSBucketChange(ctx, sink) {
```
mastermanu, line 1557: same comment applies to this enabled flag. Seems more like a toggle

app/namespace.go
```diff
@@ -1492,10 +1648,12 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 		},
 	}
 
+	export_s3_commands.Subcommands = append(export_s3_commands.Subcommands, export_general_commands...)
 	exportCommand.Subcommands = append(exportCommand.Subcommands, export_s3_commands)
 
 	// TODO: remove GCP sink feature flag check when out of private preview
```
mastermanu, line 1654: [Nit] let's update comment to reflect new nomenclature (e.g. "pre-release" instead of "private preview")

app/namespace_test.go
```diff
@@ -1656,6 +1690,254 @@ func (s *NamespaceTestSuite) TestDeleteExportSink() {
...
+						SaName:         "testSA",
+						GcpProjectName: "testGcpAccount",
+						BucketName:     "testBucket",
+						EnableCmek:     false,
+					},
+				}
+			},
+		},
+		{
+			name:         "create export sink with invalid role arn",
```
mastermanu, line 1726: [Nit] test case name seems AWS related

mastermanu (approved):

app/namespace.go
alice-yin: yes.

app/namespace.go
alice-yin: Done

app/namespace_test.go
alice-yin: yeah. The test naming is wrong.

app/namespace.go
```diff
@@ -1312,60 +1405,125 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
 					}
+					resourceVersion := c.selectExportSinkResourceVersion(ctx, sink)
+
+					isEnabledChange, err := c.isSinkEnabledChange(ctx, sink)
+					if err != nil {
+						return err
+					}
+
+					if !isEnabledChange && !c.isAssumedRoleChange(ctx, sink) && !c.isKmsArnChange(ctx, sink) && !c.isS3BucketChange(ctx, sink) {
+						fmt.Println("nothing to update")
```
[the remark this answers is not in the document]
alice-yin: No. I check the UpdateExport sink will just return empty

PR #311, opened 2024-01-19
add service account commands

mattkim:
## What was changed

add service account commands to tcld

## Why?

adding functionality for service accounts

## Checklist

1. Closes 

2. How was this tested:

unit tests

3. Any docs updates needed?


app/apikey.go
```diff
@@ -136,6 +164,7 @@ func NewAPIKeyCommand(getAPIKeyClientFn GetAPIKeyClientFn) (CommandOut, error) {
 				return err
 			},
 			Subcommands: []*cli.Command{
+				// TODO: need a create for service account command
```
tminusplus, line 167: Left over TODO? Looks like you added it in this PR

tminusplus (approved):
🚀

app/serviceaccount.go
```diff
@@ -0,0 +1,392 @@
...
+							ctx.String(serviceAccountNameFlagName),
+							ctx.String(serviceAccountDescriptionFlagName),
+							nil,
+							"",
+							nil,
+						)
+					},
+				},
+				{
+					Name:  "disable",
```
tminusplus, line 245: We de-scoped the disable functionality for the short-term, and the new proto files have an update around this. Should be able to remove it now.

app/apikey.go
tminusplus: Never mind, you added create API key but meant for service accounts.

app/serviceaccount.go
mattkim: done

PR #337, opened 2024-03-05
Support add replica in tcld

yux0:
## What was changed
Support add replica in tcld

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/namespace.go
```diff
@@ -588,6 +612,30 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 				return c.createNamespace(n, unp)
 			},
 		},
+		{
+			Name:    "add-region",
```
laniehei, line 616: imo there should be different endpoints for adding a region versus globalization. It should be intentional when a customer converts from a single region to multi-region namespace

app/namespace.go
```diff
@@ -269,6 +270,32 @@ func (c *NamespaceClient) createNamespace(n *namespace.Namespace, p []*auth.User
 	return PrintProto(res)
 }
 
+func (c *NamespaceClient) addReplica(ctx *cli.Context) error {
+	var resourceVersion string
+	if v := ctx.String(ResourceVersionFlagName); v != "" {
+		resourceVersion = v
+	}
+
+	region := ctx.String(namespaceRegionFlagName)
```
laniehei, line 279: [nit] is there a helper you could call here with predefined error?

app/namespace.go
```diff
@@ -269,6 +270,32 @@ func (c *NamespaceClient) createNamespace(n *namespace.Namespace, p []*auth.User
...
+	if len(region) == 0 {
+		return fmt.Errorf("namespace region is required")
+	}
+
+	res, err := c.client.GlobalizeNamespace(c.ctx, &namespaceservice.GlobalizeNamespaceRequest{
+		RequestId: ctx.String(RequestIDFlagName),
+		Namespace: ctx.String(NamespaceFlagName),
+		TargetRegion: &common.RegionID{
+			CloudProvider: "aws",
```
laniehei, line 288: maybe add TODO: to also require cloud when we get to it

app/namespace.go
```diff
@@ -588,6 +612,30 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
 			},
 		},
+		{
+			Name:    "add-region",
+			Usage:   "Add an new region to the temporal namespace",
+			Aliases: []string{"c"},
+			Flags: []cli.Flag{
+				RequestIDFlag,
+				ResourceVersionFlag,
+				&cli.StringFlag{
```
laniehei, line 622: [nit] isn't this defined somewhere already  where we could reuse?

laniehei (approved):
approved to not block, but have questions

app/namespace.go
```diff
@@ -269,6 +270,32 @@ func (c *NamespaceClient) createNamespace(n *namespace.Namespace, p []*auth.User
 	return PrintProto(res)
 }
 
+func (c *NamespaceClient) addReplica(ctx *cli.Context) error {
```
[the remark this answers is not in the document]
yux0: only because we are rebranding this feature.

protogen/api/common/v1/message.pb.go
```diff
@@ -24,6 +26,90 @@ var _ = math.Inf
 // proto package needs to be updated.
 const _ = proto.GoGoProtoPackageIsVersion3 // please upgrade the proto package
 
+type FailoverType int32
+
+const (
+	FAILOVER_TYPE_UNSPECIFIED FailoverType = 0
```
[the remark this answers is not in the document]
yux0: This is the generated proto code. I copy from saas-proto. It is hard to remove them from a auto generated one.

app/namespace.go
yux0: I will add a cloud provider input flag.

app/namespace.go
yux0: I will check

app/namespace.go
yux0: No in the repo right now.

PR #338, opened 2024-03-06
Add tcld namespace failover command

yux0:
## What was changed
Add tcld namespace failover command

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/namespace.go
```diff
@@ -391,6 +394,36 @@ func readAndParseCACerts(ctx *cli.Context) (read caCerts, err error) {
...
+	if len(namespace) == 0 {
+		return fmt.Errorf("namespace is required")
+	}
+
+	region := ctx.String(namespaceRegionFlagName)
+	if len(region) == 0 {
+		return fmt.Errorf("region is required")
+	}
+
+	cloudProvider := ctx.String(regionCloudProviderFlagName)
```
laniehei, line 408: we should probably just add in 'aws' as default until we're multicloud

app/namespace.go
```diff
@@ -1208,6 +1241,49 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+				},
+				&cli.StringFlag{
+					Name:  regionCloudProviderFlagName,
+					Usage: "The cloud provider of the region. Default: aws",
+					Value: "aws",
+				},
+			},
+			Action: func(ctx *cli.Context) error {
+				namespaceName := ctx.String(NamespaceFlagName)
+				region := ctx.String(namespaceRegionFlagName)
```
laniehei, line 1270: can we instead query their replicas and present them with a list of regions to fali over to?

laniehei (approved):

app/namespace.go
yux0: This is optional

app/namespace.go
yux0: We can update the behavior in the future.

app/namespace.go
```diff
@@ -391,6 +394,36 @@ func readAndParseCACerts(ctx *cli.Context) (read caCerts, err error) {
 	return parseCertificates(cert)
 }
 
+func (c *NamespaceClient) failoverNamespace(ctx *cli.Context) error {
+	namespace := ctx.String(NamespaceFlagName)
+	if len(namespace) == 0 {
+		return fmt.Errorf("namespace is required")
+	}
+
+	region := ctx.String(namespaceRegionFlagName)
```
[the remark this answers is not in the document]
yux0: for now, I want to make it explicit so customers know which region is active. I don't think it matters a lot even in a emergency case. They can look up where is the passive region.

PR #344, opened 2024-04-17
Support optional mTLS

shakeelrao:
Adds support for creating / updating namespace with different authentication methods.

### DIFF
--- a/app/namespace.go
+++ b/app/namespace.go
@@ -938,7 +938,7 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 						if err != nil {
 							return err
 						}
-						fmt.Println(authMethodToString(n.Spec.AuthMethod))
+						fmt.Println(toString(n.Spec.AuthMethod))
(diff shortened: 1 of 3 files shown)


app/namespace.go
```diff
@@ -521,6 +533,11 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 					Aliases: []string{"rd"},
 					Value:   30,
 				},
+				&cli.StringFlag{
+					Name:  authMethodFlagName,
+					Usage: "The authentication method to use for the namespace (e.g. 'mtls', 'api_key')",
+					Value: AuthMethodMTLS,
```
tminusplus, line 539: Wonder if we should default this to mTLS and make it non-required, to ensure users don't break scripts when they update tcld.

tminusplus (approved):

app/namespace.go
shakeelrao: i believe that's what `Value` does :D

PR #345, opened 2024-04-18
Add prompt to set auth method

shakeelrao:
Add a confirmation prompt to disruptive `AuthMethod` changes.

### DIFF
--- a/app/namespace.go
+++ b/app/namespace.go
@@ -1974,7 +1974,7 @@ func compareCodecSpec(existing, replacement *namespace.CodecServerPropertySpec) 
 	return diff.Diff(string(existingBytes), string(replacementBytes)), nil
 }
 
-// Determines if an `AuthMethod` update will result in dropped client connections.
+// Determines if an `AuthMethod` update will fail existing client connections.
 func disruptiveChange(old namespace.AuthMethod, new namespace.AuthMethod) bool {
 	return old != namespace.AUTH_METHOD_RESTRICTED && new != namespace.AUTH_METHOD_API_KEY_OR_MTLS
 }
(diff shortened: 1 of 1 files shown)


app/namespace.go
```diff
@@ -929,6 +929,17 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 						if n.Spec.AuthMethod == authMethod {
 							return errors.New("nothing to change")
 						}
+						if disruptiveChange(n.Spec.AuthMethod, authMethod) {
+							yes, err := ConfirmPrompt(ctx,
+								fmt.Sprintf("setting auth method to '%s' from '%s' will cause existing client connections to be dropped. "+
```
tminusplus, line 934: Could be better more specific (drop sounds temporary) like `will cause existing client connections to fail authentication`.

app/namespace.go
```diff
@@ -1963,6 +1974,15 @@ func compareCodecSpec(existing, replacement *namespace.CodecServerPropertySpec)
 	return diff.Diff(string(existingBytes), string(replacementBytes)), nil
 }
 
+// Determines if an `AuthMethod` update will result in dropped client connections.
+// Note that `old` and `new` must be different.
+func disruptiveChange(old namespace.AuthMethod, new namespace.AuthMethod) bool {
+	return (new == namespace.AUTH_METHOD_RESTRICTED) ||
```
tminusplus, line 1980: Could this be `old != namespace.AUTH_METHOD_RESTRICTED && new != old`?

tminusplus (approved):

app/namespace.go
shakeelrao: no, since `(api_key, api_key_or_mtls)` is non-disruptive.

app/namespace.go
shakeelrao: true...though the endpoint gets deleted, so client it's more of a connectivity issue rather than auth.

app/namespace.go
shakeelrao: but i think `old != namespace.AUTH_METHOD_RESTRICTED && new != namespace.AUTH_METHOD_API_KEY_OR_MTLS` works.

PR #348, opened 2024-05-28
Update region format

yux0:
## What was changed
Add region format validation

## Why?
Add region format validation

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/app/namespace.go
+++ b/app/namespace.go
@@ -15,6 +15,7 @@ import (
 	"github.com/temporalio/tcld/protogen/api/auth/v1"
 	"github.com/temporalio/tcld/protogen/api/cloud/cloudservice/v1"
 	"github.com/temporalio/tcld/protogen/api/sink/v1"
+	"github.com/temporalio/tcld/utils"
 
 	"github.com/kylelemons/godebug/diff"
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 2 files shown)


app/namespace.go
```diff
@@ -288,19 +288,18 @@ func (c *NamespaceClient) addRegion(ctx *cli.Context) error {
...
-	if len(cloudProvider) == 0 {
-		return fmt.Errorf("namespace cloud provider is required")
+	if err := utils.ValidateCloudProviderAndRegion(regionString); err != nil {
+		return err
 	}
 
 	res, err := c.cloudAPIClient.AddNamespaceRegion(c.ctx, &cloudservice.AddNamespaceRegionRequest{
 		Namespace:        ctx.String(NamespaceFlagName),
-		Region:           fmt.Sprintf("%s-%s", cloudProvider, region),
+		Region:           regionString,
```
laniehei, line 302: can we do "%s:%s" like in the other cli?

app/namespace.go
```diff
@@ -683,15 +681,10 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 				},
 				&cli.StringFlag{
 					Name:     namespaceRegionFlagName,
-					Usage:    "New region to add to the namespace.",
+					Usage:    "New region to add to the namespace. Please use format <cloud provider>-<region name>.",
```
laniehei, line 684:
```suggestion
					Usage:    "New region to add to the namespace. Please use format <cloud provider>:<region name>.",
```

app/namespace.go
```diff
@@ -1393,15 +1386,10 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 				},
 				&cli.StringFlag{
 					Name:     namespaceRegionFlagName,
-					Usage:    "The region to failover to",
+					Usage:    "The new region to failover. Please use format <cloud provider>-<region name>.",
```
laniehei, line 1389:
```suggestion
					Usage:    "The new region to failover. Please use format <cloud provider>:<region name>.",
```

yux0: This is the new cloud api format. I cannot change it.

app/namespace.go
yux0: This is the new cloud api format. I cannot change it.

app/namespace.go
```diff
@@ -298,9 +299,14 @@ func (c *NamespaceClient) addRegion(ctx *cli.Context) error {
 		return fmt.Errorf("namespace cloud provider is required")
 	}
 
+	targetRegion := fmt.Sprintf("%s-%s", cloudProvider, region)
```
jumbogo, line 302: Do we already have a cloud provider flag for tcld?

yux0: yes. I added a optional one.

laniehei (approved):

PR #350, opened 2024-06-11
Add description field to Nexus endpoint spec

nikki-dag:
https://temporalio.atlassian.net/browse/CLD-2996


app/nexus.go
```diff
@@ -97,11 +99,15 @@ func (c *NexusClient) createEndpoint(
 
 func (c *NexusClient) patchEndpoint(
 	existingEndpoint *nexus.Endpoint,
+	description string,
 	targetNamespaceID string,
 	targetTaskQueue string,
 	resourceVersion string,
 	asyncOperationId string,
 ) (*operation.AsyncOperation, error) {
+	if description != "" {
```
mastermanu, line 108: what's the tcld command to clear an existing description?

app/nexus.go
```diff
@@ -352,18 +366,20 @@ func NewNexusCommand(getNexusClientFn GetNexusClientFn) (CommandOut, error) {
 								"Only the fields that are provided will be updated",
 							Flags: []cli.Flag{
 								endpointNameFlag,
+								endpointDescriptionOptionalFlag,
```
mastermanu, line 369: do we need to update description field for command?

app/nexus.go
```diff
@@ -323,6 +335,7 @@ func NewNexusCommand(getNexusClientFn GetNexusClientFn) (CommandOut, error) {
 								"This will fail if an endpoint with the same name is already registered",
 							Flags: []cli.Flag{
 								endpointNameFlag,
+								endpointDescriptionOptionalFlag,
```
mastermanu, line 338: do we need to update description field for command?

nikki-dag: Updated description.

app/nexus.go
nikki-dag: Updated description.

app/nexus.go
nikki-dag: Added ` nexus endpoint description set/unset` commands.
Now we have 2 ways of updating the description.
- Wanted to keep it in the `update` command to be consistent with `create`
- Wanted to keep the `set` to be consistent with `unset`

Happy to change if you think one of the ways is unnecessary.

nikki-dag: Changed this too `unset-description` bool flag similar to os cli.
https://github.com/bergundy/cli/blob/187b82da3e3d8702df476bc418abf9b706f7f95a/temporalcli/commands.gen.go#L979C5-L979C56

app/nexus.go
nikki-dag: OS cli doesn't have an additional comment for the description field.
https://github.com/bergundy/cli/blob/187b82da3e3d8702df476bc418abf9b706f7f95a/temporalcli/commands.gen.go#L852

I think we can leave it for now.

app/nexus.go
```diff
@@ -267,6 +304,28 @@ func NewNexusCommand(getNexusClientFn GetNexusClientFn) (CommandOut, error) {
 		Usage:    "Namespace that is allowed to call this endpoint",
 		Required: true,
 	}
+	getEndpointDescription := func(ctx *cli.Context) (string, error) {
+		description := ctx.String(endpointDescriptionOptionalFlag.Name)
+		descriptionFile := ctx.String(endpointDescriptionFileOptionalFlag.Name)
+		if description != "" && descriptionFile != "" {
```
mastermanu, line 310: we should also guard against somebody calling unset and also setting the description

mastermanu: ignore

mastermanu (approved):

PR #355, opened 2024-06-19
Add namespace scoped service accounts

shakeelrao:
- **update to latest saas-proto**
- **impl**
- **tests**


app/account.go
```diff
@@ -61,9 +69,14 @@ func (c *AccountClient) listRegions() ([]regionInfo, error) {
 
 	var regions []regionInfo
 	for _, r := range resp.Regions {
+		rid := r.GetRegionId()
+		cp, ok := cloudProviderMap[rid.GetProvider()]
+		if !ok {
+			cp = "unspecifed"
```
jumbogo, line 75: hmm, cp="" should be good enough "unspecified" won't help much here i think

app/serviceaccount.go
```diff
@@ -210,29 +226,61 @@ func NewServiceAccountCommand(getServiceAccountClientFn GetServiceAccountClientF
...
+							if len(ctx.String(accountRoleFlagName)) == 0 {
+								return fmt.Errorf("account role must be specified; valid types are %v", accountActionGroups)
+							}
+							ag, err := toAccountActionGroup(ctx.String(accountRoleFlagName))
+							if err != nil {
+								return fmt.Errorf("failed to parse account role: %w", err)
+							}
+							acctActionGroup = ag
+						} else if scope.Type == auth.SERVICE_ACCOUNT_SCOPE_TYPE_NAMESPACE && len(ctx.String(accountRoleFlagName)) > 0 {
+							return fmt.Errorf("account role must be empty when creating a namespace scoped service account")
```
tminusplus, line 283: nit: could be "namespace scoped service accounts may not have an account role" or something similar

tminusplus (approved):

PR #367, opened 2024-07-11
Add region to create export sink

david049:
## What was changed

Adds region to creating export sink
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/namespace.go
```diff
@@ -1543,9 +1544,13 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 					}
 
 					namespace := ctx.String(NamespaceFlagName)
-					ns, err := c.getNamespace(namespace)
-					if err != nil {
-						return fmt.Errorf("unable to get namespace: %v", err)
+					region := ctx.String(RegionFlag.Name)
+					if len(region) == 0 {
```
david049, line 1548: I was thinking of having the default be `ns.Spec.Region` and check if passiveregions length > 0 to fail early, but noticed that both Region and PassiveRegions are up to be deprecated (Region deprecated in the proto, looks like there are discussions to deprecate PassiveRegions here https://github.com/temporalio/saas-proto/pull/482/files#r1640308985),

david049: We could let this go through since this will fail on the validation for create export sink call anyways

app/flags.go
```diff
@@ -55,6 +56,12 @@ var (
 		EnvVars:  []string{"TEMPORAL_CLOUD_NAMESPACE"},
 		Required: true,
 	}
+	RegionFlag = &cli.StringFlag{
+		Name:     RegionFlagName,
+		Usage:    "The region to use for the request, if not set the server will use the namespace's region",
+		Aliases:  []string{"re"},
+		Required: false,
+	}
```
david049, line 64: Should this flag be defined in the other file closer to the request instead? I wasn't sure whether the convention was to define it here or in the other file

alice-yin: Can we add this flag as sinkRegion specifically? Then it could be added to app/namespace.go

alice-yin (approved):

app/namespace.go
```diff
@@ -1535,6 +1536,12 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 					s3BucketFlagRequired,
 					RequestIDFlag,
 					kmsArnFlag,
+					&cli.StringFlag{
```
mattkim, line 1539: nit: should this be a variable to follow original design pattern

mattkim (approved):

app/namespace.go
david049: I'll make it a variable before merging

PR #390, opened 2024-09-10
add cloud provider flag for create namespace

jumbogo:
## What was changed

Add cloud provider flag to tcld namespace creation, so we can support multiple cloud.  gcp example:
```
./tcld --server "saas-api.tmprl-test.cloud:443" n c --cf cert.pem --region us-central1 --namespace zz-test-tcld-ns-gcp-2 --cloud-provider=gcp
```

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/namespace.go
```diff
@@ -2001,6 +2019,21 @@ func disruptiveChange(old namespace.AuthMethod, new namespace.AuthMethod) bool {
...
 
+func getRegionId(cloudProvider, activeRegion string) (common.RegionID, error) {
+	var regionId common.RegionID
+	switch cloudProvider {
+	case CloudProviderAWS:
+		regionId.Provider = common.CLOUD_PROVIDER_AWS
+		regionId.Name = activeRegion
+	case CloudProviderGCP:
+		regionId.Provider = common.CLOUD_PROVIDER_GCP
+		regionId.Name = activeRegion
```
alice-yin, line 2030: As regionID.Name is both activeRegion, we could take it out of the switch clause?

app/namespace.go
```diff
@@ -580,6 +581,11 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 				codecEndpointFlag,
 				codecPassAccessTokenFlag,
 				codecIncludeCredentialsFlag,
+				&cli.StringFlag{
+					Name:    cloudProviderFlagName,
+					Usage:   `Cloud provider for the namespace to be created for, currently support [aws, gcp].  For this version, if not specified, we default to aws`,
+					Aliases: []string{"cp"},
```
alice-yin, line 587: You could add a defaultValue in cli.StringFlag so you avoid the check between line 610 and 612

jumbogo: Yes, that's the "follow up" change i mentioned in the comment.  If i do this, it will be a breaking change, and some existing customer may get affected.  So would want a release of this version to let customer know about the flag, and for next release, we will enforce the flag

app/namespace.go
jumbogo: ack

app/namespace.go
alice-yin: i mean you could add a `defaultValue`  as to "aws" and that will automatic apply if user doesn't given an input.

alice-yin (approved):

app/namespace.go
jumbogo: oh, the following line "Value: CloudProviderAWS," is the default value

app/namespace.go
```diff
@@ -1946,6 +1962,20 @@ func disruptiveChange(old namespace.AuthMethod, new namespace.AuthMethod) bool {
...
 
+func getRegionId(cloudProvider, activeRegion string) (common.RegionID, error) {
+	var regionId common.RegionID
+	switch cloudProvider {
+	case CloudProviderAWS:
+		regionId.Provider = common.CLOUD_PROVIDER_AWS
+	case CloudProviderGCP:
+		regionId.Provider = common.CLOUD_PROVIDER_GCP
+	default:
+		return regionId, fmt.Errorf("unknown cloud provider specified, the supported cloud providers are aws and gcp")
```
anekkanti, line 1973: [nit]
```suggestion
		return regionId, fmt.Errorf("unknown cloud provider specified, the supported cloud providers: [%s]", strings.Join([]string{CloudProviderAWS, CloudProviderGCP}, ","))
```

app/namespace.go
```diff
@@ -1946,6 +1962,20 @@ func disruptiveChange(old namespace.AuthMethod, new namespace.AuthMethod) bool {
 	return old != namespace.AUTH_METHOD_RESTRICTED && new != namespace.AUTH_METHOD_API_KEY_OR_MTLS
 }
 
+func getRegionId(cloudProvider, activeRegion string) (common.RegionID, error) {
+	var regionId common.RegionID
+	switch cloudProvider {
```
anekkanti, line 1967: [nit]
```suggestion
	switch strings.ToLower(cloudProvider) {
```

anekkanti (approved):

PR #391, opened 2024-09-10
[Merge on 9/18] Remove nexus feature flag

nikki-dag:


app/app.go
```diff
@@ -30,16 +30,8 @@ func NewApp(params AppParams) (*cli.App, error) {
 	}
 
 	var commands []*cli.Command
+	commands = append(commands, params.Commands...)
```
mattkim, line 33: nit it seems like this can be simplified to be `app.Commands = params.Commands` ?

mattkim (approved):

PR #397, opened 2024-10-16
feat: add api key owner listing flag

### LINKED ISSUE #395: [Feature Request] Enable additional filters on GetAPIKeys
### Is your feature request related to a problem? Please describe.

We'd like to use the filtering options for getting api keys by owner ids for our ServiceAccounts that we've automated.

### Describe the solution you'd like

There are already filtering options enabled from the CloudOps API to filter api keys by owner id, so adding that as a command flag would suffice.

### Additional context

Link to the filtering options from the [protogen file](https://github.com/temporalio/tcld/blob/0314374d86777ce8e103a4b3f553113cc755a376/protogen/api/authservice/v1/request_response.pb.go#L1301-L1305)

gabizou:
fixes #395

## What was changed

This adds the `--owner-id` flag to the `tcld ak l` command, with a shorthand of `-o`.

## Why?

 Given that this will enable passing in ServiceAccount IDs or User IDs, we can better reduce the number of paginated requests being performed for automations that are centered around key rotations, while enabling continued development 

## Checklist

1. Closes #395 

2. How was this tested:

I built the command locally and ran it against our cloud instance to great success.

3. Any docs updates needed?

Updated the README for listing API keys and also took liberty to document rotating ServiceAccount keys in the same manner (since the `--si value` flag was undocumented in README).

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,12 @@ tcld apikey create --name <api-key-name> --description <api-key-description> --d
 ```
 tcld apikey list
 ```
+
+### List API Keys for a specific account (ServiceAccount or User):
+```
+tcld apikey list --owner-id <account-id>
+```
+
 ### Delete an API Key:
 ```
 tcld apikey delete --id <api-key-id>
@@ -75,6 +81,8 @@ tcld apikey enable --id <api-key-id>
 ```
 
 ### Performing an API Key rotation:
+
+#### Current User Specific Rotation
 1. Generate the new API key to rotate to.
 ```
 tcld apikey create --name <api-key-name> --description <api-key-description> --duration <api-key-duration>
@@ -82,6 +90,17 @@ tcld apikey create --name <api-key-name> --description <api-key-description> --d
 2. Update temporal clients to use the new API key and monitor deployments to make sure all old API key usage is gone.
 3. Delete the old API key.
 ``` 
+tcld apikey delete --id <api-key-id>
+```
+
+#### Service Account Specific Rotation
+1. Generate the new API key to rotate to.
+```
+tcld apikey create --name <api-key-name> --description <api-key-description> --duration <api-key-duration> --service-account-id <service-account-id>
+```
+2. Update temporal clients to use the new API key and monitor deployments to make sure all old API key usage is gone.
+3. Delete the old API key.
+```
 tcld apikey delete --id <api-key-id>
 ```
 
--- a/app/apikey.go
+++ b/app/apikey.go
@@ -12,6 +12,10 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+const (
+	ownerIDFlagName = "owner-id"
+)
+
 type (
 	APIKeyClient struct {
 		client authservice.AuthServiceClient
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -62,6 +62,12 @@ tcld apikey create --name <api-key-name> --description <api-key-description> --d
 ```
 tcld apikey list
 ```
+
+### List API Keys for a specific account (ServiceAccount or User):
```
shakeelrao, line 66: nit: `List API keys for a specific owner (service account or user)`

README.md
```diff
@@ -62,6 +62,12 @@ tcld apikey create --name <api-key-name> --description <api-key-description> --d
...
+```
+tcld apikey list --owner-id <account-id>
```
shakeelrao, line 68: nit: `--owner-id <owner-id>`

app/apikey.go
```diff
@@ -258,9 +263,17 @@ func NewAPIKeyCommand(getAPIKeyClientFn GetAPIKeyClientFn) (CommandOut, error) {
 					Name:    "list",
 					Usage:   "List apikeys",
 					Aliases: []string{"l"},
-					Flags:   []cli.Flag{},
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:    ownerIDFlagName,
+							Usage:   "The owner id of the API Keys to list",
+							Aliases: []string{"o"},
```
shakeelrao, line 270: thoughts on `oid`?

shakeelrao:
looks great -- thanks for the contribution!

app/apikey_test.go
```diff
@@ -89,6 +89,17 @@ func (s *APIKeyTestSuite) TestList() {
 	s.NoError(s.RunCmd("apikey", "list"))
 }
 
+func (s *APIKeyTestSuite) TestOwnerIdList() {
```
shakeelrao, line 92: recommend moving this logic into `TestList` above.

app/apikey.go
```diff
@@ -85,12 +89,13 @@ func (s *APIKeyClient) createServiceAccountAPIKey(
 	return PrintProto(resp)
 }
 
-func (s *APIKeyClient) listAPIKey() error {
+func (s *APIKeyClient) listAPIKey(ownerId string) error {
```
captainbeardo, line 92: nit: `ownerID`

captainbeardo (commented):
Yup, looks great.

app/apikey.go
mattkim, line 270: just double checking, this flag is optional.

not providing the owner id implies you are listing your own api keys.

mattkim (approved):

README.md
gabizou: Great note to add! I'll update

app/apikey.go
gabizou: Based on the current usage, it'll return your own api keys, or if you're a global admin, all api keys.

gabizou: And yes, this flag is always optional

README.md
gabizou: I updated with a note inside for the permissions, thoughts?

README.md
```diff
@@ -62,6 +62,14 @@ tcld apikey create --name <api-key-name> --description <api-key-description> --d
 ```
 tcld apikey list
 ```
+
+### List API keys for a specific owner (service account or user):
+Note: Only Global Admins may list API keys of other users/service
```
shakeelrao, line 67: looks good! how about simplifying to `Note: only Global Admins may list API keys of other users and service accounts.`?

README.md
shakeelrao: looks good! how about simplifying to `Note: only Global Admins may list API keys of other users and service accounts.`?

shakeelrao (approved):

Issue #349, opened 2024-06-11
[Bug] "make" fails when env TEMPORAL_CLOUD_API_KEY is set


maxramqvist:
### What are you really trying to do?

Install `tcld` according to the instructions in the docs: https://docs.temporal.io/cloud/tcld#install-tcld

### Describe the bug

If I run your instructions with the env var TEMPORAL_CLOUD_API_KEY set as an actual api key, the tests will fail:
```bash
git clone https://github.com/temporalio/tcld.git
cd tcld
make
```
Make will fail with:
```bash
FAIL	github.com/temporalio/tcld/app	10.486s
FAIL
make: *** [Makefile:27: test] Error 1
```
### Minimal Reproduction

1. Clone this repo
1. `export TEMPORAL_CLOUD_API_KEY=an-actual-api-key`
1. `make`

### Environment/Versions

This repo with commit `0549218` <- latest on main as of right now.
Go v1.22.3 on Ubuntu 22.04

briankassouf:
Fixed by #399:

```
% TEMPORAL_CLOUD_API_KEY=an-actual-api-key make
ok      github.com/temporalio/tcld/app  8.303s  coverage: 79.7% of statements
ok      github.com/temporalio/tcld/cmd/tcld     1.629s  coverage: 75.0% of statements
ok      github.com/temporalio/tcld/utils        1.243s  coverage: 64.7% of statements
passed
```
Please let us know if you continue to have issues

PR #399, opened 2024-10-17
fix bug where setting TEMPORAL_CLOUD_API_KEY interferes with unit tests

mattkim:
## What was changed

fix bug where setting TEMPORAL_CLOUD_API_KEY interferes with unit tests

## Why?

https://github.com/temporalio/tcld/issues/349


shakeelrao (approved):
seems fine, but is it possible to change the test to not rely on env variables to begin with?

PR #400, opened 2024-10-21
List available auth methods in command output

shakeelrao:
List available auth methods in the `namespace auth-method set` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,11 +45,11 @@ release:
 	$(call build,tcld,windows,amd64,.exe)
 
 tools:
-	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.2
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -45,7 +45,7 @@ release:
 	$(call build,tcld,windows,amd64,.exe)
 
 tools:
-	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.2
+	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.60.1
```
shakeelrao, line 48: upgrading to `v1.60.1` fixes linting, since `1.51.2` is incompatible with the version of go installed by GH actions.

captainbeardo (approved):

PR #402, opened 2024-11-15
using new export api for tcld

alice-yin:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


jumbogo (approved):

app/namespace.go
```diff
@@ -1469,13 +1396,14 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 				sinkNameFlag,
 			},
 			Action: func(ctx *cli.Context) error {
-				sink, err := c.getExportSink(ctx, ctx.String(NamespaceFlagName), ctx.String(sinkNameFlag.Name))
-
+				getExportSinkRes, err := c.cloudAPIClient.GetNamespaceExportSink(c.ctx, &cloudservice.GetNamespaceExportSinkRequest{
+					Namespace: ctx.String("namespace"),
```
mattkim, line 1400: nit: use const variable instead of string

app/namespace.go
```diff
@@ -1489,26 +1417,32 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 				RequestIDFlag,
 			},
 			Action: func(ctx *cli.Context) error {
-				namespaceName := ctx.String(NamespaceFlagName)
+				namespaceName := ctx.String(NamespaceFlag.Name)
 				sinkName := ctx.String(sinkNameFlag.Name)
-				resourceVersion, err := c.getExportSinkResourceVersion(ctx, namespaceName, sinkName)
```
mattkim, line 1494: are we still going to support this for older clients?

app/namespace.go
```diff
@@ -1489,26 +1417,32 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+					resourceVersion = getExportSinkRes.GetSink().GetResourceVersion()
 				}
 
-				deleteRequest := &namespaceservice.DeleteExportSinkRequest{
+				deleteRequest := &cloudservice.DeleteNamespaceExportSinkRequest{
 					Namespace:       namespaceName,
-					SinkName:        sinkName,
+					Name:            sinkName,
 					ResourceVersion: resourceVersion,
-					RequestId:       ctx.String(RequestIDFlagName),
```
mattkim, line 1503: do we not support sending the request id anymore?

app/namespace.go
alice-yin: You mean getExportSinkResourceVersion? Yes, we do. We do support all operations related to old client. But I really want to deprecate those

app/namespace.go
```diff
@@ -1521,17 +1455,15 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
 			},
 			Action: func(ctx *cli.Context) error {
-				request := &namespaceservice.ListExportSinksRequest{
-					Namespace: ctx.String(NamespaceFlagName),
-					PageSize:  int32(pageSizeFlag.Value),
-					PageToken: ctx.String(pageTokenFlag.Name),
+				request := &cloudservice.GetNamespaceExportSinksRequest{
+					Namespace: ctx.String(NamespaceFlag.Name),
+					PageSize:  int32(ctx.Int(pageSizeFlag.Name)),
+					PageToken: ctx.String("page-token"),
```
mattkim, line 1461: same comment about using const variables.

app/namespace.go
alice-yin: Yes. request id is not using any more. We're using AsyncOperationId instead

PR #403, opened 2024-11-20
[CLD-4082] update namespace scoped service account implementation

shakeelrao:
Updates the `api/auth/v1` protogen to the latest version and updates the `created-scoped` service account command to pass in the correct `ServiceAccountSpec`.


app/serviceaccount.go
```diff
@@ -307,9 +307,15 @@ func NewServiceAccountCommand(getServiceAccountClientFn GetServiceAccountClientF
...
 							Access: &auth.Access{
+								AccountAccess: &auth.AccountAccess{
+									Role: auth.ACCOUNT_ACTION_GROUP_READ,
+								},
 								NamespaceAccesses: map[string]*auth.NamespaceAccess{},
 							},
 							Description: ctx.String(serviceAccountDescriptionFlagName),
+							Scope: &auth.ServiceAccountScope{
+								Type: auth.SERVICE_ACCOUNT_SCOPE_TYPE_NAMESPACE,
+							},
```
mattkim, line 318: And we're gonna be okay if people are on the old version of tcld and get an exception right? I feel this is okay since we haven't launched the feature yet, and we can ask customers to upgrade tcld to use it.

mattkim (approved):

mattkim (approved):

app/serviceaccount.go
shakeelrao: yeah, they will get an error if they try to update a scoped service account using an older version of `tcld`, which i think is acceptable, since they need the latest version to create one.

PR #404, opened 2024-12-04
add namespace delete protection

mattkim:
## What was changed

add namespace delete protection

## Why?

give customers a way to protect namespaces from being deleted accidentally

2. How was this tested:

unit tests

3. Any docs updates needed?

will update temporal docs with new capability


app/namespace.go
```diff
@@ -638,6 +645,10 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 					}
 				}
 
+				n.Spec.Lifecycle = &namespace.LifecycleSpec{
```
shakeelrao, line 648: the default behaviour (i.e. without specifying the flag) is that delete protection is disabled, right?

app/namespace.go
```diff
@@ -668,6 +679,69 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
...
+					Usage:   "Enable delete protection on the namespace",
+					Aliases: []string{"edp"},
+				},
+				&cli.BoolFlag{
+					Name:    disableDeleteProtectionFlagName,
+					Usage:   "Enable delete protection on the namespace",
+					Aliases: []string{"ddp"},
+				},
+			},
+			Action: func(ctx *cli.Context) error {
```
shakeelrao, line 706: suggestion: we don't we follow the pattern for `auth-method` where we have a top-level NS command of `delete-protection` and two subcommands: `get` and `set`. 

https://github.com/temporalio/tcld/blob/8c55f7a12e093aed46517df0ebb90ccbafdfa3d7/app/namespace.go#L870.

app/namespace.go
mattkim: yes tested it is disabled by default

app/namespace.go
mattkim: I can add set / get.

I single flag though is a little confusing for updates because ommitting the flag is false. We would need to use strings to represent true or false.

shakeelrao: you can make the CLI flag required.

shakeelrao (approved):

PR #409, opened 2025-01-14
Update Nexus endpoint description field type

nikki-dag:


app/common.go
```diff
@@ -60,3 +63,17 @@ func parseSAPrincipal(saPrincipal string) (string, string, error) {
 
 	return saId, gcpProjectId, nil
 }
+
+func newAPIPayloadFromString(str string) *apipayload.Payload {
+	// Alternatively, use "go.temporal.io/sdk/converter" package
+	// converter.GetDefaultDataConverter().ToPayload(data)
```
nikki-dag, line 69: Ideally, we should be using the SDK converter, but we don't have a dependency on it yet.

app/connection.go
```diff
@@ -25,7 +25,7 @@ const (
 	CommitHeader                  = "tcld-commit"
 	TemporalCloudAPIVersionHeader = "temporal-cloud-api-version"
 	LegacyTemporalCloudAPIVersion = "2025-01-01-00"
-	TemporalCloudAPIVersion       = "2024-05-13-00"
+	TemporalCloudAPIVersion       = "2025-01-01-00"
```
nikki-dag, line 28: Have to ensure this won't break other commands cc @anekkanti

mastermanu (approved):

app/connection.go
anekkanti, line 28: Not in this PR, but we need to move this to v0.4.0 at some point.

anekkanti: No as far as i can tell this only impacts nexus.

anekkanti (approved):

PR #431, opened 2025-03-05
feat: allow account role to be set to none

captainbeardo:
## What was changed

If a user was added via SCIM, setting their account role to none is valid. The call will fail if the user is not a SCIM user. 

## Why?

So user's can have only group derived roles assignments. 

## Checklist

1. How was this tested:

Unit test and manual test of SCIM user

### DIFF
--- a/app/role.go
+++ b/app/role.go
@@ -15,6 +15,10 @@ var (
 	namespaceActionGroups = getNamespaceActionGroups()
 )
 
+const (
+	accountActionGroupNone = "none"
+)
+
 func getAccountActionGroups() []string {
 	var rv []string
 	for n, v := range auth.AccountActionGroup_value {
@@ -63,7 +67,10 @@ func getNamespaceRolesBatch(ctx context.Context, client authservice.AuthServiceC
 	return res.Roles, nil
 }
 
-func getAccountRole(ctx context.Context, client authservice.AuthServiceClient, actionGroup string) (*auth.Role, error) {
+func getAccountRole(ctx context.Context, client authservice.AuthServiceClient, actionGroup string, allowNone bool) (*auth.Role, error) {
+	if allowNone && strings.ToLower(strings.TrimSpace(actionGroup)) == accountActionGroupNone {
+		return nil, nil
+	}
 	ag, err := toAccountActionGroup(actionGroup)
 	if err != nil {
 		return nil, err
--- a/app/user.go
+++ b/app/user.go
@@ -146,7 +146,7 @@ func (c *UserClient) inviteUsers(
 	var roleIDs []string
 
 	// first get the required account role
-	role, err := getAccountRole(c.ctx, c.client, accountRole)
+	role, err := getAccountRole(c.ctx, c.client, accountRole, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


shakeelrao:
does this need to be exposed via `tcld`? curious what the use-case is.

app/role.go
```diff
@@ -63,7 +67,10 @@ func getNamespaceRolesBatch(ctx context.Context, client authservice.AuthServiceC
 	return res.Roles, nil
 }
 
-func getAccountRole(ctx context.Context, client authservice.AuthServiceClient, actionGroup string) (*auth.Role, error) {
+func getAccountRole(ctx context.Context, client authservice.AuthServiceClient, actionGroup string, allowNone bool) (*auth.Role, error) {
+	if allowNone && strings.ToLower(strings.TrimSpace(actionGroup)) == accountActionGroupNone {
```
shakeelrao, line 71: suggestion: i'd rename the `allowNone` to `allowUnspecified` and remain consistent to with the enum definition.

app/user.go
```diff
@@ -257,11 +257,21 @@ func (c *UserClient) setAccountRole(
 		return err
 	}
 	var newRoleIDs []string
-	accountRoleToSet, err := getAccountRole(c.ctx, c.client, accountRole)
+	accountRoleToSet, err := getAccountRole(c.ctx, c.client, accountRole, true)
 	if err != nil {
 		return err
 	}
-	if accountRoleToSet.Spec.AccountRole.ActionGroup == auth.ACCOUNT_ACTION_GROUP_ADMIN {
+	if accountRoleToSet == nil {
```
shakeelrao, line 264: if possible, i'd recommend having `getAccountRole` return `auth.ACCOUNT_ACTION_GROUP_UNSPECIFIED` instead of `nil`.

app/role.go
captainbeardo: Unspecified is different than none. In this case the user has no actual account role, it's not that they have an account role of unspecified.

app/user.go
```diff
@@ -146,7 +146,7 @@ func (c *UserClient) inviteUsers(
 	var roleIDs []string
 
 	// first get the required account role
-	role, err := getAccountRole(c.ctx, c.client, accountRole)
+	role, err := getAccountRole(c.ctx, c.client, accountRole, false)
```
mattkim, line 149: would this throw a nil pointer exception below? 

`roleIDs = append(roleIDs, role.GetId())`

app/user.go
```diff
@@ -146,7 +146,7 @@ func (c *UserClient) inviteUsers(
 	var roleIDs []string
```
mattkim, line 146: just so i understand--we are allowing a state where we can set account role to nil--and namespace roles to not nil?

mattkim: Also do we need to update the access package validation to allow the account role to be empty?

captainbeardo: Yes, account role can be nil, and namespace roles can be empty or not empty. 

The access validation already supports no account role if the user is a SCIM user.

app/user.go
captainbeardo: This doesn't allow nil to be returned because none is invalid for inviting a user. It'd get an error.

PR #444, opened 2025-04-29
feat: Metrics read role for service accounts

captainbeardo:
## What was changed

Adds the metrics read role for service accounts.

## Why?

Upcoming role allowed for service accounts.


app/role.go
```diff
@@ -11,11 +11,24 @@ import (
...
+	allAccountActionGroups  = getAllAccountActionGroups()
+	namespaceActionGroups   = getNamespaceActionGroups()
 )
 
-func getAccountActionGroups() []string {
+func getUserAccountActionGroups() []string {
+	var rv []string
+	for n, v := range auth.AccountActionGroup_value {
+		if v != int32(auth.ACCOUNT_ACTION_GROUP_UNSPECIFIED) &&
+			v != int32(auth.ACCOUNT_ACTION_GROUP_METRICS_READ) {
```
mattkim, line 23: can a user not have the metrics read role?

captainbeardo: No, we prevent that and this prevents tcld from trying.

PR #445, opened 2025-05-05
Update protogen

pglass:
## What was changed

Update protogen, to include latest cloud protos.

## Checklist

1. Closes 

2. How was this tested:

   * Follow up PR: https://github.com/temporalio/tcld/pull/446

3. Any docs updates needed?


hehaifengcn (approved):
seems changes contains other updates unrelated to s2c. Assume these are expected.

captainbeardo (approved):
I think the large number of changes is because people would cherry-pick the changes they need from the protos, and not commit all of the changes if it didn't affect what they were working on. This is probably fine though. There are a couple of things that are being exposed but not released, like metrics read role, but if they go to the trouble of picking the generated code out of tcld, maybe they deserve access to it.

PR #446, opened 2025-05-05
Add 'tcld migration' commands

pglass:
## What was changed

Add `tcld migration {get,list,start,handover,confirm,abort}` commands for self-hosted to cloud migration.

- [ ] Depends on https://github.com/temporalio/tcld/pull/445

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


hehaifengcn:
with this change, s2c api will be available public in next `tcld` release? are we ready for it?

captainbeardo (approved):
Looks good. I wouldn't mind if you added something to the README since that ends up being what informs the official tcld docs. Is this available for everybody or do they need to go through support to enable migrations on an account/namespace? I echo Haifeng's question, are we ready for that?

PR #448, opened 2025-05-12
feat: don't error on no changes, just log

captainbeardo:
## What was changed

For users and service accounts, check the error and if it is because of no changes, log the status, but error. This leaves the exit code 0 so automation tools can proceed. 

## Why?

Automation will stop if there are errors with no changes, this makes the operation idempotent. 

## Checklist

1. Closes 

2. How was this tested:

Unit tests and manual user updates.

3. Any docs updates needed?

### DIFF
--- a/app/common.go
+++ b/app/common.go
@@ -48,7 +48,7 @@ func isNothingChangedErr(ctx *cli.Context, e error) bool {
 	if !ok {
 		return false
 	}
-	return s.Code() == codes.InvalidArgument && s.Message() == "nothing to change"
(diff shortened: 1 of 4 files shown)


app/serviceaccount.go
```diff
@@ -183,6 +183,10 @@ func (c *ServiceAccountClient) performUpdate(
 
 	resp, err := c.client.UpdateServiceAccount(c.ctx, req)
 	if err != nil {
+		if isNothingChangedErr(err) {
+			fmt.Println("nothing to change")
```
mattkim, line 187: remove println?

mattkim (approved):

shakeelrao (approved):

PR #449, opened 2025-06-05
add connectivity rule operation

jumbogo:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -28,8 +28,17 @@ func NewApp(params AppParams) (*cli.App, error) {
 			InsecureConnectionFlag,
 			EnableDebugLogsFlag,
 		},
-		Commands: params.Commands,
 	}
-
+	var commands []*cli.Command
+	for _, command := range params.Commands {
+		if command.Name == "connectivity-rule" {
+			if IsFeatureEnabled(ConnectivityRuleFeatureFlag) {
+				commands = append(commands, command)
+			}
+		} else {
+			commands = append(commands, command)
+		}
+	}
+	app.Commands = commands
 	return app, nil
 }
--- a/app/feature.go
+++ b/app/feature.go
@@ -16,7 +16,7 @@ var (
 	featureflagFileName         = "feature.json"
 )
 
-var supportFeatureFlags = []string{GCPSinkFeatureFlag}
+var supportFeatureFlags = []string{GCPSinkFeatureFlag, ConnectivityRuleFeatureFlag}
 
 type FeatureFlag struct {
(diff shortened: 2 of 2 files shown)


app/connectivity_rule.go
```diff
@@ -0,0 +1,208 @@
...
+		Name:     regionFlagName,
+		Aliases:  []string{"r"},
+		Usage:    "The region of the connection",
+		Required: false,
+	}
+	cloudProviderFlag := &cli.StringFlag{
+		Name:     cloudProviderFlagName,
+		Aliases:  []string{"cp"},
+		Usage:    "The cloud provider of the connection, currently only support 'aws' and 'gcp'",
+		Required: false,
```
alice-yin, line 100: This could set as required field?

app/connectivity_rule.go
```diff
@@ -0,0 +1,208 @@
...
+		Name:     isPrivateFlagName,
+		Aliases:  []string{"ip"},
+		Usage:    "Whether or not the connectivity rule is private, if specified true, it's a private connectivity rule, and required to have connection id, region, cloud provider, gcp project id (if it's gcp) specified.  If specified false or not set, it's a public connectivity rule",
+		Required: false,
+	}
+	regionFlag := &cli.StringFlag{
+		Name:     regionFlagName,
+		Aliases:  []string{"r"},
+		Usage:    "The region of the connection",
+		Required: false,
```
alice-yin, line 94: same here

app/connectivity_rule.go
alice-yin: oh. I see. only neeed for isPrivateFlagName

alice-yin (commented):
As this is a pre-release feature, you don't want all people could access this cmd in tcld. What extra you need to add is using IsFeatureEnabled in tcld. You could find some code change around it.

app/connectivity_rule.go
jumbogo: see above

app/connectivity_rule.go
```diff
@@ -0,0 +1,208 @@
...
+			Name:    "connectivity-rule",
+			Aliases: []string{"cr"},
+			Before: func(ctx *cli.Context) error {
+				var err error
+				c, err = getConnectivityRuleClientFn(ctx)
+				return err
+			},
+			Subcommands: []*cli.Command{
+				{
+					Name:        "create",
```
anekkanti, line 125: No `delete`?

PR #451, opened 2025-06-09
Add user group commands to tcld CLD-5021

captainbeardo:
## What was changed

This adds group related commands to support native user groups. 

## Why?

So users don't have to use APIs directly. 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


app/user_group.go
```diff
@@ -222,6 +222,149 @@ func (c *UserGroupClient) setAccess(ctx *cli.Context, groupID string,
 	return PrintProto(resp.GetAsyncOperation())
 }
 
+// createGroup creates a new user group with the specified display name and account role
+func (c *UserGroupClient) createGroup(_ *cli.Context, displayName string, accountRole string) error {
+	aRole := accountRoleToAccess(accountRole)
```
briankassouf, line 227: What about namespace roles?

app/user_group.go
```diff
@@ -222,6 +222,149 @@ func (c *UserGroupClient) setAccess(ctx *cli.Context, groupID string,
...
+			},
+			AsyncOperationId: ctx.String(RequestIDFlagName),
+		}
+
+		resp, err := c.client.RemoveUserGroupMember(c.ctx, req)
+		if err != nil {
+			return fmt.Errorf("unable to remove user %s from group: %w", email, err)
+		}
+
+		if err := PrintProto(resp.GetAsyncOperation()); err != nil {
```
briankassouf, line 319: Will it be confusing if we print multiple async operation objects?

app/user_group.go
```diff
@@ -222,6 +222,149 @@ func (c *UserGroupClient) setAccess(ctx *cli.Context, groupID string,
...
+			return err
+		}
+	}
+	return nil
+}
+
+// listGroupMembers lists the members of a group using the cloud services API
+func (c *UserGroupClient) listGroupMembers(
+	pageToken string,
```
briankassouf, line 328: I see other list commands doing the pagination automatically: https://github.com/temporalio/tcld/blob/ea55673852b7ab54458e30c4b4922b8352f34f6d/app/namespace.go#L310C27-L327

Do we want to do the same here?

PR #459, opened 2025-07-08
Support detaching all namespace connectivity rules

cludden:
related-conversation: https://temporaltechnologies.slack.com/archives/C07GT2LR2TV/p1752001979253649

## What was changed
Support detaching all namespace connectivity rules

## Why?
related-conversation: https://temporaltechnologies.slack.com/archives/C07GT2LR2TV/p1752001979253649

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/app/namespace.go
+++ b/app/namespace.go
@@ -2142,6 +2142,10 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 		Flags: []cli.Flag{
 			NamespaceFlag,
 			connectivityRuleIdsFlag,
+			&cli.BoolFlag{
+				Name:  "remove-all",
+				Usage: "Acknowledge that all connectivity rules will be removed, enabling connectivity from any source",
+			},
 		},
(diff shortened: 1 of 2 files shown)


alice-yin (approved):

david049 (approved):

app/namespace.go
```diff
@@ -2142,15 +2142,22 @@ func NewNamespaceCommand(getNamespaceClientFn GetNamespaceClientFn) (CommandOut,
 		Flags: []cli.Flag{
 			NamespaceFlag,
 			connectivityRuleIdsFlag,
+			&cli.BoolFlag{
+				Name:  "remove-all",
```
david049, line 2146: small nit: do we want to define this at the top of file with the other flags

cludden: Since it's a local flag specific to this nested subcommand, I'm comfortable keeping it closer to the definition (as opposed to ~2k lines above) and it looks like there is precedent in this same file for doing so.

PR #476, opened 2025-08-30
Max duration is 2 years, so make example 1.5 years

bechols:
## What was changed
Change example string from 2.5y to 1.5y

## Why?
Because max API key duration is 2 years

## Checklist

1. How was this tested:
Ran locally, confirmed output is fixed and that apikey create command works the same as it did before

2. Any docs updates needed?
No


shakeelrao (approved):
heh...

PR #477, opened 2025-09-22
Support removing all client CA certs from a namespace

shakeelrao:


captainbeardo (approved):
Maybe add some tests?

PR #491, opened 2026-02-10
Switch from gogo to protoc-gen-go, fix build

chndler:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


anekkanti (dismissed):

go.mod
```diff
@@ -3,21 +3,24 @@ module github.com/temporalio/tcld
 go 1.23.0
 
 require (
+	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.36.11-20260209202127-80ab13bee0bf.1
 	github.com/go-playground/validator/v10 v10.13.0
 	github.com/gogo/protobuf v1.3.2
```
anekkanti: I would have expected this to go away.

anekkanti: I think we should remove all references to gogo.

chndler: Updated

app/print.go
```diff
@@ -26,19 +27,29 @@ func PrintObj(i interface{}) error {
...
+			Indent:          "\t",
+			EmitUnpopulated: true,
+		}
+		b, err := marshaler.Marshal(pm)
+		if err != nil {
+			return "", err
+		}
+		return string(b), nil
+	}
+	b, err := json.MarshalIndent(m, "", "\t")
```
anekkanti, line 45: Will the protojson produce the similar output as jsonpb? 
If not this backward incompatible and effects all outputs by tcld.

